PR #6, opened 2026-07-27
AGE-1379: add sub-agent completion tool.response in agentthreadcontext

thesujai:
> [!NOTE]
> **Medium Risk**
> Changes turn event persistence and context-append wiring for all tool calls; incorrect handling could cause missing or duplicate tool.response in replay/API consumers.
> 
> **Overview**
> **`tool.response` events are now written to the turn event store** when they stream out of execution, instead of being treated as stream-only like model deltas. **`AgentThread`** still yields each `tool.response` for live consumers, but **`appendToContext` no longer puts those events in `AGENT_CONTEXT_APPEND.output`**—only the tool-role context rows are appended—so persistence happens once at the session/`TurnHandle` layer.
> 
> Model message / delta handling is split out explicitly: they remain **stream-only**, with durable assistant content still coming from **`AGENT_CONTEXT_APPEND.output`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2906ad9a38f7800485dab9663f25e2f6c3507d90. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/harness/src/agent-session/TurnHandle.ts
+++ b/packages/harness/src/agent-session/TurnHandle.ts
@@ -328,8 +328,14 @@ export class TurnHandle<TTurnCustom extends object = Record<string, never>> {
     switch (event.type) {
       case HarnessEventType.MODEL_MESSAGE:
       case HarnessEventType.MODEL_MESSAGE_DELTA:
+        // Stream only — durable model content lands via AGENT_CONTEXT_APPEND.output.
+        return event;
+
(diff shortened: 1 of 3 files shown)


chiragjn (commented):
discussed offline - we will now start returning actual subagent tool call output instead of empty string and yield from send

chiragjn (approved):

PR #12, opened 2026-07-30
Implement re subscribe to stream

sr07asthana:
> [!NOTE]
> **Medium Risk**
> New live-stream path touches turn execution SSE, Redis TTL semantics, and a breaking SDK shape for `TurnStateDone.output`; misconfigured TTLs or multi-replica in-memory mode could drop resumes.
> 
> **Overview**
> **Resumable turn streaming** is the main change: create-turn now **dual-writes** each event into an `EventSubscriptionRegistry` (Redis streams when `REDIS_URL` is set, otherwise a shared in-memory store) before writing SSE, and a new **GET** `/api/v1/sessions/{sessionId}/turns/{turnId}/subscribe` endpoint replays from `after_sequence_number` or `Last-Event-ID`, returns **412** when the live stream is gone, and stops at `turn.done` with configurable stream/subscribe TTLs.
> 
> Harness/API contract fixes ride along: **`TurnStateDone.output`** is required but **nullable** again (OpenAPI + `TurnHandle`), **`ResponseFormat`** and **`AgentOutputEvent`** use named/discriminated Zod unions for Fern codegen, and the **TypeScript SDK** is regenerated (Fern bump) with **`sessions.subscribeToTurn`** and resumable SSE client support.
> 
> Supporting work: **`test:eventsub`** contract tests for both backends, `.env.example` docs for turn-stream env vars, and an **AGENTS.md** rule preferring `z.discriminatedUnion` when a shared discriminator exists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b95e57e7e45d8c6caf44bba37ffd3e1d4de919c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
+import type { RedisClientType } from 'redis';
+import { InMemoryEventStreamStore, InMemoryEventSubscription } from './inMemory';
+import { RedisEventSubscription } from './redis';
+
+export interface EventSubscriptionPutOptions {
+  /** Sets the lifetime of the whole stream after this append. */
+  streamTTLSeconds?: number | undefined;
+}
+
+export type SequencedEvent<T extends object> = T & { sequence_number: number };
```
debajyoti-truefoundry, line 10: What if `sequence_number` is already part of `T`

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  constructor(
+    private readonly redis: RedisClientType | undefined,
+    private readonly parseEvent: (raw: unknown) => T,
+  ) {}
+
+  get(streamId: string): EventSubscription<T> {
+    if (this.redis) {
+      return new RedisEventSubscription(this.redis, streamId, this.parseEvent);
+    }
+    this.memoryStore ??= new InMemoryEventStreamStore<T>();
```
debajyoti-truefoundry: Keep this global.

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    return new InMemoryEventSubscription(this.memoryStore, streamId);
+  }
+}
+
+/** Redis/in-memory key for one turn's resumable event stream. */
+export function turnStreamId(tenantId: string, sessionId: string, turnId: string): string {
```
debajyoti-truefoundry: move to caller.

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  return `agent:turn:${tenantId}:${sessionId}:${turnId}:stream`;
+}
+
+/** Reject new subscriptions when the stream expires within this window. */
+export const SUBSCRIBE_STREAM_THRESHOLD_MS = 60 * 1_000;
+/** Delay between poll iterations while a stream has no new events. */
+export const SUBSCRIBE_STREAM_POLL_SLEEP_INTERVAL_MS = 1_000;
+
+export function sleep(ms: number): Promise<void> {
```
debajyoti-truefoundry: util.

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/** Redis/in-memory key for one turn's resumable event stream. */
+export function turnStreamId(tenantId: string, sessionId: string, turnId: string): string {
+  return `agent:turn:${tenantId}:${sessionId}:${turnId}:stream`;
+}
+
+/** Reject new subscriptions when the stream expires within this window. */
+export const SUBSCRIBE_STREAM_THRESHOLD_MS = 60 * 1_000;
+/** Delay between poll iterations while a stream has no new events. */
+export const SUBSCRIBE_STREAM_POLL_SLEEP_INTERVAL_MS = 1_000;
```
debajyoti-truefoundry: Is this not Redis-specific? Can we not use node stream for in memory?

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  constructor(readonly streamId: string) {
+    super(`Cannot read from stream, stream does not exist anymore: ${streamId}`);
+    this.name = 'StreamGoneError';
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+/** The stream expires too soon to accept a new subscription (map to HTTP 412). */
+export class StreamExpiringError extends Error {
```
debajyoti-truefoundry: Do not need this.

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+export const SUBSCRIBE_STREAM_THRESHOLD_MS = 60 * 1_000;
+/** Delay between poll iterations while a stream has no new events. */
+export const SUBSCRIBE_STREAM_POLL_SLEEP_INTERVAL_MS = 1_000;
+
+export function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+/** The stream expired or was never created (map to HTTP 412). */
+export class StreamGoneError extends Error {
```
debajyoti-truefoundry, line 61: should this be implements or extends?

packages/server/src/runtime/event-subscription/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    readonly expiresAtMs: number,
+  ) {
+    super(`Cannot subscribe to stream, stream is about to expire: ${streamId}`);
+    this.name = 'StreamExpiringError';
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+/** A stored entry violates the event-subscription wire format (map to HTTP 412). */
+export class StreamCorruptEntryError extends Error {
```
debajyoti-truefoundry: If this is 5xx, i will just throw a normal error.

packages/server/src/runtime/event-subscription/inMemory.ts
```diff
@@ -0,0 +1,110 @@
...
+      throw new StreamExpiringError(this.streamId, stream.expiresAtMs);
+    }
+
+    let cursor = afterSequenceNumber === undefined ? 0 : afterSequenceNumber + 1;
+    for (;;) {
+      const live = this.store.getLiveStream(this.streamId);
+      if (!live) {
+        throw new StreamGoneError(this.streamId);
+      }
+      const batch = live.events.slice(cursor);
```
debajyoti-truefoundry, line 146: Can we use node streams here.

packages/server/src/runtime/event-subscription/redis.ts
```diff
@@ -0,0 +1,101 @@
...
+ * One Redis Stream with locally minted sequence numbers. Assumes a single
+ * producer per stream; the counter dies with the instance, the data via key TTL.
+ */
+export class RedisEventSubscription<T extends object> implements EventSubscription<T> {
+  private nextSequenceNumber = 0;
+
+  constructor(
+    private readonly redis: RedisClientType,
+    private readonly streamId: string,
+    private readonly parseEvent: (raw: unknown) => T,
```
debajyoti-truefoundry: Not sure if this is worth it at this point.
by default let us do type assertion, customer can update it if they want.

packages/server/src/apis/turns.ts
```diff
@@ -277,13 +328,81 @@ export function createTurnsRouter(deps: TurnsRouterDeps) {
...
+    if (!session) {
+      return c.json({ error: { message: `Session not found: ${sessionId}` } }, 404);
+    }
+    const turn = await session.getTurn(turnId);
+    if (!turn) {
+      return c.json({ error: { message: `Turn not found: ${turnId}` } }, 404);
+    }
+
+    // Server-side subscribe timeout; the SSE loop checks the signal between events.
+    const timeoutAbortController = new AbortController();
```
debajyoti-truefoundry: What if the client disconnected?

packages/server/src/apis/turns.ts
```diff
@@ -277,13 +328,81 @@ export function createTurnsRouter(deps: TurnsRouterDeps) {
     });
   };
 
+  const subscribeTurnHandler: RouteHandler<typeof subscribeTurnRoute> = async c => {
+    const { sessionId, turnId } = c.req.valid('param');
+    const body = c.req.valid('json');
+    const afterSequenceNumber = resolveAfterSequenceNumber(c, body.after_sequence_number);
```
debajyoti-truefoundry: Can we just support header? cc @chiragjn

packages/server/src/apis/turns.ts
```diff
@@ -277,13 +328,81 @@ export function createTurnsRouter(deps: TurnsRouterDeps) {
...
+
+    const session = await deps.sessions.get({ tenant_id: TENANT_ID, session_id: sessionId });
+    if (!session) {
+      return c.json({ error: { message: `Session not found: ${sessionId}` } }, 404);
+    }
+    const turn = await session.getTurn(turnId);
```
debajyoti-truefoundry, line 350: In turn handle, can we add an async static method to get turn given session and turn id?

Also, check if the session id is globally unique today?

packages/server/src/apis/turns.ts
```diff
@@ -277,13 +328,81 @@ export function createTurnsRouter(deps: TurnsRouterDeps) {
...
+        turnId,
+        afterSequenceNumber,
+        timeoutMs,
+      });
+      timeoutAbortController.abort(new Error('subscribe-timeout'));
+    }, timeoutMs);
+
+    const generator = deps.eventSubscriptions.get(turnStreamId(TENANT_ID, sessionId, turnId)).poll(afterSequenceNumber);
+
+    // Pre-flight: force the generator's stream checks (gone/expiring/corrupt)
```
debajyoti-truefoundry: Why don't we have a probe function that checks all this?

packages/server/src/apis/turns.ts
```diff
@@ -277,13 +328,81 @@ export function createTurnsRouter(deps: TurnsRouterDeps) {
...
+    // to run before SSE headers are sent, so they can still map to HTTP 412.
+    let iterResult: IteratorResult<SequencedEvent<StoredTurnStreamingEvent>, void>;
+    try {
+      iterResult = await generator.next();
+    } catch (error) {
+      clearTimeout(timeoutHandler);
+      if (
+        error instanceof StreamGoneError ||
+        error instanceof StreamExpiringError ||
+        error instanceof StreamCorruptEntryError
```
debajyoti-truefoundry: going away.

packages/sdk/src/api/resources/sessions/client/Client.ts
```diff
@@ -1113,4 +1113,126 @@ export class SessionsClient {
             },
         });
     }
+
+    /**
+     * Subscribe to the live SSE stream for a turn. Pass `after_sequence_number` to resume after a disconnect (exclusive — events after this sequence number are replayed).
+     */
+    public subscribeToARunningTurn(
```
debajyoti-truefoundry: Can we just call this subscribe turn or something?

packages/server/src/runtime/event-subscription/index.ts
sr07asthana: The harness's TurnStreamingEvent members don't carry sequence_number,
also an intersection doesn't error on duplicate keys it takes an intersection of the property types.

packages/server/src/runtime/event-subscription/index.ts
sr07asthana: the server constructs exactly one registry. it's per registry instance. All create-turn and subscribe handlers go through that single instance. so we can keep a single readonly memoryStore

packages/server/src/runtime/event-subscription/index.ts
sr07asthana: it should stay extends Error. implements only enforces a compile-time shape , the class wouldn't actually be an Error at runtime, so it would have no stack trace and error instanceof StreamGoneError / instanceof Error checks

packages/server/src/apis/turns.ts
sr07asthana: will be taken up as a separate ticket. https://linear.app/truefoundry/issue/AGE-1504/add-a-static-way-to-create-in-turnhandle-when-turn-id-session-id-and

packages/server/src/runtime/event-subscription/inMemory.ts
sr07asthana: For our backend we need a log array that is replayable (a subscriber can start at any after_sequence_number, including one from a connection that died minutes ago). A node streams Readable buffer is destroyed as it's read. We will need to keep separate array for storing replay events and add stream objects on top: one Readable created per subscriber

sr07asthana: The subscription is a replayable multi-reader log (resume from any cursor, N concurrent subscribers, TTL expiry), which a Node stream doesn't model — a Readable is a single-consumer, consume-once pipe, so we'd still keep the array/registry/TTL code and add per-subscriber stream lifecycle on top. EventEmitter + once(emitter, streamId, { signal }) is the same wake-up primitive streams use internally, giving instant delivery and abort-safe teardown with three lines and no buffer duplication.

packages/server/src/runtime/event-subscription/redis.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  async *poll(
+    afterSequenceNumber?: number,
+    options?: EventSubscriptionPollOptions,
+  ): AsyncGenerator<SequencedEvent<T>, void, unknown> {
+    const signal = options?.signal;
+    let cursor = afterSequenceNumber === undefined ? '0-0' : `${String(afterSequenceNumber + 1)}-0`;
+    for (;;) {
```
chiragjn, line 71: why are we not breaking on `turn.done` ?

chiragjn (approved):

PR #14, opened 2026-07-30
serve FE from server

pranjaltrvd:
> [!NOTE]
> **High Risk**
> Breaking change for all API consumers (`/v1` → `/api/v1`) plus new static-serving and routing behavior on the same port; deployment and SDK clients must be updated together.
> 
> **Overview**
> **Single-process deployments** now expose the chat UI and API on one origin (e.g. `:8790`): `/api/*` and `/healthz` hit the server, other GET/HEAD paths are served from `FRONTEND_DIR` (precompressed assets where present, on-the-fly gzip for Monaco workers). Without a build at that path, startup logs a warning and the process stays API-only—intended for `pnpm dev` with Vite on `:3000`.
> 
> **Docker / Compose** collapse to one root `Dockerfile` image: a `frontend-builder` stage produces `packages/frontend/dist`, copied into the runner with `FRONTEND_DIR` set. The `frontend` service, Caddyfile, and `packages/frontend/Dockerfile` are removed; the server healthcheck uses **`GET /healthz`** instead of `/`.
> 
> **Breaking API surface**: every harness route, OpenAPI/Swagger, and generated **`packages/sdk`** paths move from `/v1/...` to **`/api/v1/...`**. Docs live at `/api/v1/docs` and `/api/v1/openapi.json`.
> 
> **Frontend wiring**: catalog calls use `/api/v1/*`; gateway SDK clients get **`harnessFetch`**, which rewrites `/v1/agents/(draft-)?sessions` to `/api/v1/sessions` (replacing Vite path rewrites). Dev adds `pnpm dev` (API + Vite) and `pnpm dev:server:ui`; Vite proxies only `/api`. Production builds emit `.br`/`.gz` via `vite-plugin-compression2`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33dbe00df543327f281c24cc48d6f0260aab3481. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docker-compose.yml
```diff
@@ -55,14 +56,14 @@ services:
         condition: service_healthy
       redis:
         condition: service_healthy
-    # Probes existing GET / (plain text OK!); node is already in the image.
+    # Probes GET /healthz (plain text OK!); node is already in the image.
     healthcheck:
       test:
```
chiragjn, line 60: We should use docker compose's native url healthchecks instead of cmd like this

packages/server/package.json
```diff
@@ -12,9 +12,10 @@
     "migrate": "tsx --env-file=.env src/db/postgres/migrate-cli.ts",
     "typecheck": "tsc --noEmit",
     "openapi:write": "tsx --env-file=.env.test scripts/write-openapi.ts",
-    "test": "tsx --env-file=.env.test --test src/**/*.test.ts",
+    "test": "tsx --env-file=.env.test --test \"src/**/*.test.ts\"",
     "test:store": "jest --config jest.store.config.cjs",
-    "start": "node dist/main.js"
+    "start": "node dist/main.js",
+    "preview": "node --env-file=.env dist/main.js"
```
chiragjn: does `dev` not work for our use case

packages/server/Dockerfile
```diff
@@ -40,6 +40,27 @@ COPY packages/harness packages/harness
...
+COPY patches patches
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm fetch
+
+COPY package.json .npmrc ./
+COPY packages/harness/package.json packages/harness/package.json
+COPY packages/server/package.json packages/server/package.json
+COPY packages/frontend/package.json packages/frontend/package.json
+# Same as builder: the root postinstall hook needs the build:gen inputs.
+COPY packages/harness/scripts packages/harness/scripts
+COPY packages/harness/src/core/sandbox/scripts packages/harness/src/core/sandbox/scripts
```
chiragjn: It should be possible to collapse all of these into fewer COPY commands

packages/frontend/src/harnessFetch.ts
chiragjn, line 1: we will move all BE apis behind /api/ prefix and only rely on that instead of going deeper

packages/server/src/app.ts
```diff
@@ -47,13 +48,15 @@ export interface ServerDeps {
   redis: RedisClientType;
   /** Request-reply dispatch table served by this replica's executor. */
   requestReplyRouter: RequestReplyRouter;
+  /** Frontend build to serve; omitted or absent = API only. */
+  frontendDir?: string;
```
chiragjn: Let's start simple, let's just assume FE is always there and let's not give this flexibility right now
dev | start command will always build FE first and then start the server

later we will give an api only mode for server

packages/server/src/frontend.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+import { serveStatic } from '@hono/node-server/serve-static';
+import type { MiddlewareHandler } from 'hono';
+import { every } from 'hono/combine';
+import { compress } from 'hono/compress';
+import { existsSync, readFileSync } from 'node:fs';
+import path from 'node:path';
+
+/** Routes the server answers itself; never served from the build. */
+const SERVER_PATH_PREFIXES = ['/v1', '/docs', '/openapi.json', '/healthz'];
```
chiragjn: we will only have /api and /healthz

packages/server/src/frontend.ts
```diff
@@ -0,0 +1,73 @@
...
+/** Browser navigation accepts HTML; a missing script does not, and gets the JSON 404. */
+export function wantsAppShell(request: { method: string; pathname: string; accept: string | undefined }): boolean {
+  if (request.method !== 'GET' && request.method !== 'HEAD') return false;
+  if (isServerPath(request.pathname)) return false;
+  return request.accept?.toLowerCase().includes('text/html') ?? false;
+}
+
+export function loadFrontendAssets(dir: string): FrontendAssets | undefined {
+  const indexPath = path.join(dir, 'index.html');
+  if (!existsSync(indexPath)) return undefined;
```
chiragjn: let it throw and 404

packages/server/src/frontend.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  // serveStatic resolves `root` against the working directory; absolute paths are outside its contract.
+  const root = path.relative(process.cwd(), dir) || '.';
```
chiragjn:
```
import { fileURLToPath } from 'node:url'
import { serveStatic } from '@hono/node-server/serve-static'

app.use(
  '/static/*',
  serveStatic({ root: fileURLToPath(new URL('./', import.meta.url)) })
)
```


The root option resolves paths relative to the current working directory (process.cwd()). This means the behavior depends on where you run your Node.js process from, not where your source file is located. If you start your server from a different directory, file resolution may fail.

For reliable path resolution that always points to the same directory as your source file, use import.meta.url:

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
```
debajyoti-truefoundry, line 1: We should move this dockerfile to root level.

debajyoti-truefoundry: Outside of server.

packages/frontend/tests/harnessFetch.test.ts
```diff
@@ -0,0 +1,75 @@
+import assert from 'node:assert/strict';
```
debajyoti-truefoundry, line 1: Please create a test folder in the same structure as the src folder and keep tests there.

package.json
```diff
@@ -12,8 +12,7 @@
     "format:fix": "prettier --write \"**/*.{js,jsx,ts,tsx,json,md}\" --cache --config .prettierrc",
     "test": "pnpm --filter @truefoundry/utils test && pnpm --filter @truefoundry/server test && pnpm --filter frontend test",
     "test:store:local": "bash scripts/test-store-with-postgres.sh",
-    "dev:server": "pnpm --filter @truefoundry/server dev",
-    "preview": "pnpm build && pnpm --filter @truefoundry/server preview",
+    "dev:server": "pnpm --filter frontend build && pnpm --filter @truefoundry/server dev",
```
chiragjn: We don't need to build FE in dev mode, right?

chiragjn: I was thinking is it possible

```
dev:server  // only BE, does not re-build frontend. Maybe we can add an arg for re-building 
dev:frontend // only FE
dev // starts both in dev mode
```

packages/frontend/vite.config.ts
```diff
@@ -49,7 +49,7 @@ export default defineConfig({
     port: 3000,
```
chiragjn: We should make this configurable

packages/server/src/frontend.test.ts
chiragjn, line 1: tests should be `server/tests/`

packages/server/src/frontend.ts
chiragjn, line 1: what does "shell" mean in this file, bit confusing

packages/server/src/frontend.ts
```diff
@@ -0,0 +1,87 @@
...
+function isServerPath(pathname: string): boolean {
+  return SERVER_PATH_PREFIXES.some(prefix => pathname === prefix || pathname.startsWith(`${prefix}/`));
+}
+
+/** Only a browser navigating to a client-side route gets the shell; scripts and fetches keep their 404. */
+function wantsAppShell(c: Context): boolean {
+  if (c.req.method !== 'GET' && c.req.method !== 'HEAD') return false;
+  if (isServerPath(c.req.path)) return false;
+  return c.req.header('accept')?.toLowerCase().includes('text/html') ?? false;
+}
```
chiragjn: does this mean any xhr for frontend js, css files etc will fail?

packages/server/src/frontend.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (c.req.method !== 'GET' && c.req.method !== 'HEAD') return next();
+    return serveCompressed(c, next);
+  };
+
+  app.use('/*', serveBuild);
+
+  // Vary because compress() above may encode the shell without advertising it.
+  const shellHeaders = { 'Cache-Control': REVALIDATE_CACHE_CONTROL, Vary: 'Accept-Encoding' };
+
+  app.notFound(c => (wantsAppShell(c) ? c.html(readShell(), 200, shellHeaders) : routeNotFound(c)));
```
chiragjn: So I am guessing we fall through here if path is not found in FE folder
but I am not getting why we need `wantsAppShell(c) ? c.html(readShell(), 200, shellHeaders)` and not direct 404

can we please add some explanation as a comment

chiragjn (approved):
Keeping it approved. let's resolve comments and merge

PR #18, opened 2026-07-31
Add single-replica mode so Redis becomes optional.

pranjaltrvd:
> [!NOTE]
> **Medium Risk**
> Changes turn-cancel routing and deployment defaults; wrong SINGLE_BINARY/REDIS pairing can leave cross-replica cancels ineffective, though behavior is documented.
> 
> **Overview**
> Adds **`SINGLE_BINARY`** (default **`true`**) so a one-process deployment can run **without Redis**: **`REDIS_URL`** is ignored, **`EXECUTOR_ID`** is fixed to **`local`**, and startup skips Redis and the request-reply executor.
> 
> For multi-replica setups, set **`SINGLE_BINARY=false`** and provide **`REDIS_URL`**; **`docker-compose.yml`** keeps peering enabled that way for smoke tests. **`cancelSessionTurn`** only hops to a peer when a Redis client exists; otherwise it stays local (so misconfigured multi-replica can still fail to cancel remotely).
> 
> Docs/env (**.env.example**, **.env.test**), OpenAPI generation, and unit tests were updated; server tests live under **`tests/unit/`** with separate TS projects from DB tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ca53376185e5d1b4b22032e50b547ce55873fa6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docker-compose.yml
```diff
@@ -40,6 +40,10 @@ services:
       PORT: 8790
       # Compose network hostname; credentials still come from env_file.
       POSTGRES_HOST: postgres
+      # Compose runs one replica but keeps peering on so the peered cancel path
+      # stays exercised in development. A single-replica deployment can set this
+      # to true and drop the redis service entirely.
+      SINGLE_REPLICA: 'false'
```
debajyoti-truefoundry: Call it a single binary mode.

chiragjn (approved):

PR #21, opened 2026-08-01
Add session deletion API

bhaveshpatel640:
Add delete session API with full store cleanup

---

> [!NOTE]
> **Medium Risk**
> Permanent data deletion across session persistence layers; concurrent delete vs. active turn/SSE needs correct behavior, which this PR tests but production Postgres may depend on how related rows are removed when only the `session` row is deleted.
> 
> **Overview**
> Adds **`DELETE /api/v1/sessions/{sessionId}`** (OpenAPI + route), returning **204** and removing the session through **`ISessionStore.deleteSession`**. Deletes are **tenant-scoped**, **idempotent** when the session is already gone, and the in-memory store explicitly drops turns and event logs; SQLite backends rely on **ON DELETE CASCADE** from the `session` row.
> 
> Harness **store contract tests** now cover full cleanup (no orphaned context/events on ID reuse), post-delete mutation failures, tenant mismatch no-ops, prefix session IDs, and races with concurrent appends. A few “missing turn” expectations now use **`TurnNotFoundError`** instead of **`SessionStoreNotFoundError`**.
> 
> After deletion, session/turn HTTP APIs surface **404** (new integration test). Turn **SSE** treats mid-stream removal as a **warn**-level **`SessionStoreNotFoundError`** instead of an unexpected error. Postgres/SQLite store comments are updated to clarify that multiple **`running`** turns are allowed when not chaining off a still-running predecessor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b74e42048ab1f8e801985a28dfd82aa4fdb357b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/server/src/db/postgres/session-store/queries/sessions.ts
```diff
@@ -91,6 +92,47 @@ export async function createSession(db: Kysely<Database>, input: CreateSessionIn
...
+
+    await trx
+      .selectFrom('turn')
+      .select('turn_id')
+      .where('tenant_id', '=', input.tenant_id)
+      .where('session_id', '=', input.session_id)
+      .forUpdate()
+      .execute();
+
+    for (const table of [
```
debajyoti-truefoundry: Please reduce the number of DB calls.

fern/openapi/openapi.json
```diff
@@ -2762,6 +2762,36 @@
       }
```
debajyoti-truefoundry, line 3152: Some general comments.
1. Will a FK and cascade delete make things easier? Will FK affect any postgres ingestion rate?  Will it be fine if we have proper indexes? Let us say we have 100 inserts per second for now.
2. Should we just use FK cascade in SQLite if it's supported?
3. Have we ensured that we cannot append any turn-related data when the session is deleted? How are we achieving this? Will (1) make things simple?
4. How are we doing error handling if a session is deleted while we are appending turn data.

bhaveshpatel640: > Will a FK and cascade delete make things easier? Will FK affect any postgres ingestion rate? Will it be fine if we have proper indexes? Let us say we have 100 inserts per second for now.

Yes.

- FK + `ON DELETE CASCADE` makes deletion one parent delete, dependent rows gets deleted with it.
- Postgres: 8 queries reduced to 3 (session lock, turn locks, parent delete).
- At ~100 inserts/sec, FK cost is negligible: insert checks the session PK, cascade uses child indexes already leading with `(tenant_id, session_id, …)`.

--

> Should we just use FK cascade in SQLite if it's supported?

Yes.

- SQLite supports `ON DELETE CASCADE`, when client has `PRAGMA foreign_keys = ON` (We already have it enabled)
- SQLite cannot `ALTER` add FKs, so the migration rebuilds child tables with cascade. Delete is then one parent delete.
- Docs: [SQLite ALTER TABLE](https://www.sqlite.org/lang_altertable.html)

--

> Have we ensured that we cannot append any turn-related data when the session is deleted? How are we achieving this? Will (1) make things simple?

Yes.

- Children FK to `session (tenant_id, session_id)`, after delete, new rows are rejected
- Postgres locks the session and its turns before delete
- Predelete writes are cascaded away; post-delete writes fail because the parent is gone

--

> How are we doing error handling if a session is deleted while we are appending turn data.

- Delete cancels turns (peers best-effort), cascade-wipes the session, then aborts local runs.
- Mid stream writes after delete fail in the store, FKs will keep the DB clean.

packages/server/src/db/postgres/session-store/queries/sessions.ts
bhaveshpatel640: Yes, Done

fern/openapi/openapi.json
debajyoti-truefoundry, line 3149: 1. Let us keep the session ID globally unique.
2. Let all the child tables will have session ID field only, remove tenant ID. FK will be session ID. Cascade Delete.

debajyoti-truefoundry: Ensure we update the indexes correctly.

debajyoti-truefoundry: Let us ensure we get the indexes reviewed.

debajyoti-truefoundry: We cannot cancel all the turns as of now.

We can delete everything and fail if someone is trying to append anything in the child table.

packages/harness/tests/agent-session/store/storeContractSuite.ts
```diff
@@ -144,6 +144,307 @@ export function runStoreContractSuite(createStore: () => ISessionStore) {
...
+          store.patchThreadCapabilityState({
+            ...keys,
+            thread_id: MAIN_THREAD_ID,
+            key: 'tfy.plan',
+            state: { step: 1 },
+          }),
+      ];
+
+      for (const write of writes) {
+        await expect(write()).rejects.toBeInstanceOf(SessionStoreNotFoundError);
```
debajyoti-truefoundry: Should it not be SessionNotFound.

debajyoti-truefoundry: Should this not be turn not found?

packages/harness/tests/agent-session/store/storeContractSuite.ts
```diff
@@ -144,6 +144,307 @@ export function runStoreContractSuite(createStore: () => ISessionStore) {
...
+            turn_id: 'turn-1',
+            limit: 10,
+            page_token: undefined,
+            order: 'asc',
+          })
+        ).data,
+      ).toEqual([]);
+    });
+
+    it('rejects all session and turn mutations after deleteSession', async () => {
```
debajyoti-truefoundry: session mutation, turn mutation, event mutation.

packages/harness/tests/agent-session/store/storeContractSuite.ts
```diff
@@ -144,6 +144,307 @@ export function runStoreContractSuite(createStore: () => ISessionStore) {
     });
   });
 
+  describe('deleteSession', () => {
+    it('removes all session data, is idempotent, and preserves tenant isolation', async () => {
```
debajyoti-truefoundry: The deletion is no-op if tenant id not matching.

packages/harness/tests/agent-session/store/storeContractSuite.ts
bhaveshpatel640: Yes, updated the new and existing tests

packages/server/src/db/postgres/session-store/queries/sessions.ts
```diff
@@ -91,6 +92,30 @@ export async function createSession(db: Kysely<Database>, input: CreateSessionIn
...
+ */
+export async function deleteSession(db: Kysely<Database>, input: DeleteSessionInput): Promise<void> {
+  await db.transaction().execute(async trx => {
+    await trx
+      .selectFrom('turn as t')
+      .innerJoin('session as s', 's.session_id', 't.session_id')
+      .select('t.turn_id')
+      .where('s.tenant_id', '=', input.tenant_id)
+      .where('t.session_id', '=', input.session_id)
+      .forUpdate('t')
```
debajyoti-truefoundry: Why do we need this? 

Is this not DB's responsibility?

debajyoti-truefoundry (approved):

packages/server/src/db/postgres/session-store/queries/sessions.ts
bhaveshpatel640: Removed the turn FOR UPDATE

PR #55, opened 2026-08-05
Add Together AI as a provider, with 20 verified models

pranjaltrvd:
Adds `together` as a provider type, with the 20 Together models that actually work on a plain API key.

## Adapter

Together stays on the shared OpenAI-compatible adapter rather than getting its own case. Its official `@ai-sdk/togetherai` package sends `response_format: {"type":"json_object"}` and drops the schema outright, warning `unsupported` — the same regression that keeps Fireworks off its package. Captured side by side on `Qwen/Qwen3.7-Plus`:

| Adapter | `response_format` on the wire |
| --- | --- |
| `@ai-sdk/togetherai` | `{"type":"json_object"}` |
| compatible | `{"type":"json_schema","json_schema":{…,"strict":true}}` |

The package exposes no provider options of its own, so there is nothing to trade for that loss.

## Which models

`truefoundry/models` has 375 Together entries; 94 are chat models that are not deprecated and declare `function_calling`. Together serves only 20 of those without a dedicated endpoint:

- 48 fail with `Unable to access non-serverless model` — they exist only as dedicated deployments.
- 4 more claim `provisioning: serverless` in the repo but are rejected too, so the flag cannot be trusted on its own.

The catalog therefore lists what answered a live tool call, not what the metadata promised.

## Verification

- 20/20 catalog models pass `tool-single`.
- All 10 advertised reasoning efforts across 4 models were accepted, including `max`, which travels as `xhigh`.
- Full 30-scenario suite over 4 curated models: 81 pass. One real quirk found — Together's Kimi endpoint requires `top_p=0.95` and rejects anything else, which Moonshot's own API does not. Four rate-limit errors passed on a serial re-run. Both recorded in the provider quirks doc.
- Driven end to end through the server: tool call, reasoning and signature replay on a follow-up turn, on four Together models.

The UI picks the new type up automatically, since the provider form reads the generated SDK enums.


packages/sdk/.fern/metadata.json
```diff
@@ -27,10 +27,9 @@
         "streamType": "web",
         "useDefaultRequestParameterValues": true
     },
-    "originGitCommit": "2fa4d1229e085b8dab98f45717ebd82db9311e5c",
+    "originGitCommit": "300657be94fe920fc4d849b12f3ee8540a195207",
     "originGitCommitIsDirty": true,
-    "invokedBy": "ci",
+    "invokedBy": "manual",
     "requestedVersion": "0.0.0",
```
chiragjn: did we update openapi and sdk offline?

packages/harness/src/core/llm/VercelAILLM.ts
```diff
@@ -174,6 +175,7 @@ export function buildLanguageModel(config: VercelAIProviderConfig): LanguageMode
     }
     case 'fireworks':
     case 'zai':
+    case 'together':
```
chiragjn, line 179: let's please add a comment fo which of these vercel sdks are available and why are we not using them

packages/harness/src/core/llm/VercelAILLM.ts
```diff
@@ -174,6 +175,7 @@ export function buildLanguageModel(config: VercelAIProviderConfig): LanguageMode
     }
     case 'fireworks':
```
chiragjn, line 177: the fireworks sdk seems to do some handling for reasoning and thinking on top of openai compatible sdk

https://github.com/vercel/ai/blob/fbb154a00008375c46b7bcc1a6dde8ae5eef2f8d/packages/fireworks/src/fireworks-provider.ts#L144-L197

are we then doing this handling correctly?

packages/harness/src/core/llm/VercelAILLM.ts
```diff
@@ -174,6 +175,7 @@ export function buildLanguageModel(config: VercelAIProviderConfig): LanguageMode
...
     case 'zai':
```
chiragjn, line 178: https://github.com/vercel/ai/pull/17340/changes zai is also upcoming
we should keep a TODO

packages/harness/src/core/llm/VercelAILLM.ts
```diff
@@ -101,9 +102,9 @@ export interface VercelAILLMConfig {
 /**
  * Every OpenAI-compatible provider shares this adapter and differs only by endpoint, which the
  * caller resolves. The provider name doubles as the `providerOptions` key, which is why
- * {@link buildProviderOptions} can key on it directly. Fireworks and Z AI stay here on purpose:
- * `@ai-sdk/fireworks` downgrades `json_schema` to a schema-less `json_object`, and the only Z AI
- * package is a community one that drops replayed reasoning.
+ * {@link buildProviderOptions} can key on it directly. Fireworks, Together and Z AI stay here on
+ * purpose: the Fireworks and Together packages both downgrade `json_schema` to a schema-less
```
chiragjn: how did we verify this?
toegether seems to mostly passthrough to openai compatible
https://github.com/vercel/ai/blob/fbb154a00008375c46b7bcc1a6dde8ae5eef2f8d/packages/togetherai/src/togetherai-provider.ts#L157-L161

packages/server/src/schemas/modelProvider.ts
```diff
@@ -43,6 +44,7 @@ export const PROVIDER_DEFAULT_BASE_URLS: Partial<Record<ProviderType, string>> =
   fireworks: 'https://api.fireworks.ai/inference/v1',
   zai: 'https://api.z.ai/api/paas/v4',
   moonshot: 'https://api.moonshot.ai/v1',
+  together: 'https://api.together.xyz/v1',
 };
```
chiragjn, line 48: I have created another ticket to refactor well known, call supplied

We should create 1 zod schema for each provider and base url should go as default value in each schema

chiragjn (commented):
Added comments

packages/harness/src/core/llm/VercelAILLM.ts
chiragjn, line 602: why not equal alibaba?
we need some comments

chiragjn, line 1: for anything mapping to openai compatible, we can reflect get a list of standard keys like 
`service_tier`, `thinking`, `cache_control`, `thinking_config`, `thinking_budget`, `prompt_cache_key`
if these are present then we can pass them along in provideroptions

packages/harness/src/core/llm/VercelAILLM.ts
```diff
@@ -401,13 +418,10 @@ function moonshotProviderOptions({
...
   reasoningEffort: string | undefined;
 }): JSONObject | undefined {
-  // Typed against the package so a change to the literal it accepts breaks the build, not the call.
-  const effort: Pick<MoonshotAIProviderOptions, 'reasoningEffort'> =
-    reasoningEffort === 'max' ? { reasoningEffort: 'max' } : {};
   const thinking = readBodyField({ rawBody, key: 'thinking' });
   const reasoningHistory = readBodyField({ rawBody, key: 'reasoning_history' });
   const opts: JSONObject = {
-    ...effort,
+    ...maxEffortOption(reasoningEffort),
```
chiragjn, line 420: let's comment why we are doing this

chiragjn (approved):
Keeping it approved

PR #62, opened 2026-08-05
Curate the model catalog down to current frontier models

pranjaltrvd:
## What

Trims the shipped model catalog from 231 presets to 46. The catalog is discovery-only — it feeds `GET /api/v1/settings/model-providers/catalog` so the settings UI can offer presets to copy into a `PUT` body — but the presets it offered had accumulated every model each provider has ever exposed, including many that cannot drive an agent's tool-calling loop.

The rule applied per provider: keep the current frontier generation plus the immediately preceding flagship as a fallback. Dropped are superseded generations, dated snapshots that duplicate a live alias, floating `-latest` aliases (they make agent behaviour non-reproducible when the provider reroutes them), and small or task-specific models.

## Per provider

- **OpenAI** — the GPT-5.6 family (Sol, Terra, Luna) is the current line and OpenAI lists it as the replacement for essentially everything older; `gpt-5`, `o3` and the Codex/chat-latest snapshots already have shutdown dates. Kept the 5.6 trio, `gpt-5.5`, `gpt-5.5-pro`, and `gpt-5.4-mini` as a non-frontier small tier. `gpt-5.5-pro` stays because 5.6 has no separate pro model — pro is a reasoning *mode* on Sol, which `reasoning_efforts` cannot express.
- **Anthropic** — current lineup is Fable 5, Opus 5, Sonnet 5 and Haiku 4.5; Opus 4.8/4.7/4.6 and Sonnet 4.6/4.5 are documented as legacy. Kept the four current models plus `claude-opus-4-8` and `claude-sonnet-4-6`.
- **Gemini** — the Pro and Flash lines have desynced. Flash reached 3.6, but the Pro line is still at Gemini 3.1 Pro and there is no 3.5 or 3.6 Pro, so `gemini-3.1-pro-preview` remains the flagship entry. Kept it plus 3.6 Flash, 3.5 Flash and 3.5 Flash-Lite; dropped the 2.5 series, Gemma and the robotics models.
- **Fireworks / Together** — kept the frontier open-weight tier (DeepSeek V4 Pro and Flash, GLM-5.2, Kimi K3, MiniMax M3) plus Kimi K2.7-Code and the Fireworks `-fast` routers. Dropped gpt-oss, Llama 3.3, Gemma, Qwen2.5, Nemotron and the superseded K2.6 / MiniMax M2.7. Together also gains the provider logo every other entry carries, and its model names lose the duplicated vendor prefix.
- **Z.ai** — GLM-5.2, 5.1 and 5-Turbo (the last is tuned for long agent tool-calling chains).
- **Moonshot** — K3 and the two K2.7-Code variants.
- **Alibaba** — `qwen3.8-max` (the new 2.4T flagship), the 3.7 max/plus/flash tier, and the third-party frontier models Alibaba hosts.

The same open-weight models appear under several providers on purpose, since a tenant may hold only one provider's key.

## Notes

`properties` blocks are byte-identical to before for every retained model; this change only removes entries. One pre-existing gap it makes more visible: `claude-opus-5` and `claude-sonnet-5` carry no `reasoning_efforts`, so the settings UI likely will not offer effort control on Anthropic's two newest models. Left for a follow-up.

## Testing

`pnpm --filter @truefoundry/server test` — 96 tests across 16 suites pass. The catalog tests iterate over whatever the YAML ships (asserting every preset is configurable via `PUT`), so they cover the reduced set. No code references the removed model IDs; the only match outside the YAML was a docstring example (`openai/gpt-5-6-sol`), which is retained.


chiragjn (approved):
@shubhamagarwal-create to let us know if we need to keep some of these models

packages/server/src/catalog/model-catalog.yaml
chiragjn, line 1: for together do we want to keep all models from here?
https://docs.together.ai/docs/serverless/models#chat-models

PR #64, opened 2026-08-05
Limit well-known model provider types to one configured provider

pranjaltrvd:
## Why

The settings UI keys its provider cards by type, so a tenant holding two providers of the same well-known type renders inconsistently. A well-known type also has a fixed endpoint — the server already knows where to send requests — so a second row of that type is only ever a duplicate credential. Rejecting it costs nothing real and keeps the UI in a state it can render.

## What

`PUT /api/v1/settings/model-providers` now answers **409** when the incoming type is well-known and another provider of that type is already configured under a different name.

The guard engages only for the seven well-known types (`openai`, `anthropic`, `google-gemini`, `fireworks`, `zai`, `moonshot`, `together`). It reuses the existing `WELL_KNOWN_BASE_URL_TYPES` list in `schemas/modelProvider.ts` via a new `isWellKnownProviderType` predicate, so the two groupings cannot drift apart.

Behaviour that is explicitly preserved:

- **Updating the configured provider.** The upsert stays keyed by `name`, and the check skips a row with the same name, so replacing models or rotating the API key works exactly as before.
- **Caller-supplied types stay unrestricted.** `alibaba` and `custom` both require the request to carry its own `base_url` — an Alibaba MaaS host embeds the workspace id, and `custom` is arbitrary by definition. Several rows of those types are genuinely different endpoints, so capping them at one would block a real use case.

## Status code choice

409 rather than the 422 this repo uses elsewhere for semantic rejections. Here 422 means "the body is valid but references a resource this server does not provide"; this is instead a conflict with existing state, which is what 409 is for. The status is declared on the route, so the generate-sdk workflow will regenerate `fern/openapi/openapi.json` and `packages/sdk` on this PR — both generated artifacts are left untouched in the commit.

## Known limits

The check is application-level, so two concurrent first writes could still both land. Closing that would need a unique partial index over the manifest's `type` in both the Postgres and SQLite migrations, which is more than this interim limit warrants.

The frontend is unchanged: the UI surfaces the 409 message when someone attempts a duplicate rather than hiding the option up front. If we later want it never offered, the place is `getModelProviderCatalog` in `packages/frontend/src/modelProviderCatalog.ts`, filtering out well-known types that `listModelProviders` already returns.

## Testing

`pnpm --filter @truefoundry/server test` — 99 tests across 16 suites pass, including three new cases: the duplicate rejection (asserting nothing partially landed, via both the provider list and the `GET /models` read view), the same-name update path, and two `custom` providers coexisting. `pnpm typecheck`, `pnpm lint:ci` and `pnpm format:check` are clean.

One note on the new test: routers are exercised in isolation here, without the app's `onError`, which is what wraps an `HTTPException` message into the `{ error: { message } }` JSON envelope. The test therefore asserts on the plain-text body, matching how the rest of the suite treats router-level errors.


packages/server/src/schemas/modelProvider.ts
```diff
@@ -105,26 +70,83 @@ const ModelProviderManifestBaseSchema = z
...
+  OpenAiModelProviderSchema,
+  AnthropicModelProviderSchema,
+  GoogleGeminiModelProviderSchema,
+  FireworksModelProviderSchema,
+  ZaiModelProviderSchema,
+  MoonshotModelProviderSchema,
+  TogetherModelProviderSchema,
+  AlibabaModelProviderSchema,
+  CustomModelProviderSchema,
+] as const;
```
chiragjn, line 149: let's just inline this in z.discriminatedUnion

packages/server/src/schemas/modelProvider.ts
```diff
@@ -105,26 +70,83 @@ const ModelProviderManifestBaseSchema = z
...
+function wellKnownProviderSchema<Type extends Exclude<ProviderType, 'custom'>>({
+  type,
+  base_url,
+}: {
+  type: Type;
+  base_url: string;
+}) {
+  return ModelProviderManifestBaseSchema.extend({
+    type: z.literal(type),
+    name: z.literal(type).default(type),
```
chiragjn, line 86:
```suggestion
```

packages/server/src/schemas/modelProvider.ts
chiragjn, line 1: What do we call the DTO entry vs the manifest type?

ideally it should be something like
ListModelProvidersRequest -> 

```
ListModelProvidersResponse {
  data: ModelProvider[]
}

ModelProvider {
id: ...
manifest: ModelProviderManifest
}
```

packages/server/src/db/sqlite/model-provider-store/SqliteModelProviderStore.ts
```diff
@@ -52,7 +52,7 @@ export class SqliteModelProviderStore implements IModelProviderStore {
       .insertInto('model_provider')
       .values({
         tenant_id: input.tenant_id,
-        name: input.name,
+        name: input.manifest.name,
```
chiragjn, line 55: for custom input.manifest.name,
for non custom input.manifest.type

packages/server/src/db/postgres/model-provider-store/PostgresModelProviderStore.ts
```diff
@@ -52,7 +52,7 @@ export class PostgresModelProviderStore implements IModelProviderStore {
       .insertInto('model_provider')
       .values({
         tenant_id: input.tenant_id,
-        name: input.name,
+        name: input.manifest.name,
```
chiragjn, line 55: same here

packages/frontend/src/modelProviderCatalog.ts
```diff
@@ -45,10 +45,12 @@ export function toHarnessModelEntry(model: UiModelEntry): Harness.ModelEntry {
 }
 
 export function toUiModelProvider(provider: Harness.ModelProvider): UiModelProvider {
+  // Every type but `custom` is named after itself, so the wire leaves `name` optional.
+  const name = provider.name ?? provider.type;
```
chiragjn, line 49: this will be based on provider.type
custom -> provider.name 
non custom -> provider.type

chiragjn (approved):
Keeping this approved

PR #69, opened 2026-08-06
AGE-1622: add admin and user claim config

heerambavi1998:
> [!NOTE]
> **Medium Risk**
> Changes how admin access is determined from OIDC tokens; misconfiguration could deny admin or default users to non-admin, though behavior is isolated in new code with tests and not yet shown wired into the callback in this diff.
> 
> **Overview**
> Adds **configurable OIDC ID-token claims** so deployments can map IdP-specific identity and admin membership without hardcoding Okta-style `groups`.
> 
> **Configuration:** When OIDC is enabled, `OIDC_USER_REFERENCE_CLAIM` (default `sub`), `OIDC_USER_ROLE_CLAIM` (default `groups`), and `OIDC_ADMIN_ROLE_VALUE` (default `admin`, case-sensitive) are loaded into `OIDCConfig` and documented in `.env.example`.
> 
> **Claims module:** New `auth/claims` helpers derive `UserContext` (`userRef` + `admin`/`user`) from decoded tokens—missing/invalid reference claims throw; malformed role claims downgrade to `user` instead of crashing. `buildAuthorizationRequestParams` builds scopes plus essential `id_token` claims (and adds the `groups` scope when the role claim is `groups` for Okta).
> 
> Unit tests cover claim normalization, role resolution (arrays vs bare strings, Azure-style `roles`), and authorization request shaping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 745c023da2080a917634ec02b8b88e9cfd5e5d9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/server/src/config.ts
```diff
@@ -219,7 +223,18 @@ function resolveOIDCConfig(): OIDCConfig | undefined {
...
+    OIDC_CLIENT_ID: clientId,
+    OIDC_CLIENT_SECRET: clientSecret,
+    OIDC_USER_REFERENCE_CLAIM:
+      getEnv('OIDC_USER_REFERENCE_CLAIM', { defaultValue: DEFAULT_OIDC_USER_REFERENCE_CLAIM }) ??
+      DEFAULT_OIDC_USER_REFERENCE_CLAIM,
+    OIDC_USER_ROLE_CLAIM:
+      getEnv('OIDC_USER_ROLE_CLAIM', { defaultValue: DEFAULT_OIDC_USER_ROLE_CLAIM }) ?? DEFAULT_OIDC_USER_ROLE_CLAIM,
+    OIDC_ADMIN_ROLE_VALUE:
+      getEnv('OIDC_ADMIN_ROLE_VALUE', { defaultValue: DEFAULT_OIDC_ADMIN_ROLE_VALUE }) ?? DEFAULT_OIDC_ADMIN_ROLE_VALUE,
+  };
```
thesujai, line 237: Already your are marking the defaultValue:, we need a fallback? with ??

packages/server/src/auth/claims.ts
```diff
@@ -0,0 +1,93 @@
+import type { OIDCConfig } from '../config';
+
+/** Raw claims from a decoded ID token; values are untyped until read here. */
+export type IdTokenClaims = Record<string, unknown>;
+
+export type Role = 'admin' | 'user';
```
thesujai, line 6: Can be enum

thesujai (approved):

packages/server/src/config.ts
heerambavi1998: Yes, because getEnv returns optionally. Same elsewhere

packages/server/src/auth/claims.ts
heerambavi1998: We don't use ts enums anywhere, no strong reason to use here

PR #79, opened 2026-08-06
AGE-1506: add created by to session

heerambavi1998:
> [!NOTE]
> **Medium Risk**
> Requires a DB migration and makes `created_by` a required field on Session API responses; creator identity is not wired to real auth yet, so filtering behavior may not match production expectations until that lands.
> 
> **Overview**
> Adds **immutable `created_by`** on sessions end-to-end: harness `SessionRecord`/schemas, in-memory and Postgres/SQLite stores, OpenAPI, and session list/get/create responses.
> 
> **`GET /api/v1/sessions`** accepts optional **`created_by`** (same pattern as `agent_id`). Postgres and SQLite migrations add a non-null `created_by` column (default `''` for existing rows).
> 
> Session **create** requires `created_by` at the store layer; the HTTP API currently sets **`created_by: 'trueforge-default'`** on every new session rather than deriving it from the authenticated caller. Create request body is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af55ce9d68a3b2e0ab09ba1b664c6dae88a7a57a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #84, opened 2026-08-06
Split model providers into an identity DTO and a manifest

pranjaltrvd:
Follow-up to the review comments on #79, which merged before they were addressed.

### What changed

A provider's name is not configuration. Every type but `custom` is named after its own `type`, so the manifest now carries configuration alone and the DTO pairs it with the derived name:

```
PUT  /settings/model-providers   body: ModelProviderManifest
GET  /settings/model-providers   -> { data: ModelProvider[] }

ModelProvider { name, manifest: ModelProviderManifest }
```

Well-known manifests no longer accept a `name` at all — passing one is a 400 — while `custom` still requires one, since it is the only type that supplies its own endpoint. The name is derived in one place, `modelProviderName`, which the API layer calls before handing the row's key to the store. Both stores went back to a plain `name: input.name`, so neither backend needs to understand the provider union, and `UpsertProviderInput` now matches skills and MCP servers.

### Comment by comment

1. `MODEL_PROVIDER_SCHEMAS` is inlined in `z.discriminatedUnion` — it had one caller.
2. `name` is gone from the well-known manifests, which removes the `z.literal(type).default(type)` pinning trick.
3. Manifest and DTO are now distinct: `ModelProviderManifest` is the document and PUT body, `ModelProvider` is the wire entry. Per-type OpenAPI schemas were renamed to match (`AnthropicModelProviderManifest`, and so on).
4/5. The `custom ? name : type` rule lives in `modelProviderName` rather than being duplicated in the SQLite and Postgres stores.
6. The frontend needs no `type` branch: it reads `provider.name` from the DTO and takes configuration from `provider.manifest`.

### Notes

- No data migration: nothing parses the stored jsonb on read, so rows written since #79 that still carry a `name` keep working; dev databases can just be recreated.
- `openapi.json` and the SDK were regenerated with the workflow's own steps and committed separately, because GitHub had no hosted runners available.
- Verified locally end to end: workspace typecheck and lint clean, 313 SDK tests, 156 server unit tests, 30 frontend tests, and the store contract suites against both Postgres and SQLite.

---

> [!NOTE]
> **Medium Risk**
> This is an API contract change: clients that PUT or expect `name` on well-known providers will get 400s or type mismatches until updated. Persistence behavior is unchanged for existing JSONB rows that still contain a redundant `name`.
> 
> **Overview**
> **Well-known model providers no longer accept or return a `name` field** — identity is implied by `type` (e.g. `openai`), while **`custom` still requires a caller-chosen `name`**. Sending `name` on a built-in type is now rejected at validation.
> 
> The server derives the database upsert key in one place via **`modelProviderName`**: the PUT handler passes that `name` into the store instead of reading `manifest.name`. SQLite/Postgres upserts use **`input.name`** only, so stores no longer interpret the provider union.
> 
> OpenAPI, SDK types/serialization, and route descriptions are regenerated/updated to match. The frontend maps list cards by **`type` for well-known providers** and **`name` for custom**, with a new unit test for custom IDs.
> 
> Catalog docs now say presets become PUT bodies after dropping catalog-only **`logo`/`name`** and adding **`auth`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05f5f29c50751828f3d5a1c96f6628153045eea1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fern/openapi/openapi.json
```diff
@@ -194,7 +194,7 @@
         ],
         "description": "Agent Definition"
       },
-      "AlibabaModelProvider": {
+      "AlibabaModelProviderManifest": {
```
debajyoti-truefoundry: I would avoid creating a concept of manifest now, as it is not there anywhere else.

Please keep the structure same and just remove the `name`.

If we need to duplicate the types, so be it for now.

debajyoti-truefoundry: > I would avoid creating a concept of manifest now

In the public API.
We can take this call globally later.

debajyoti-truefoundry:
```
PUT  /settings/model-providers   body: ModelProviderManifest
GET  /settings/model-providers   -> { data: ModelProvider[] }

ModelProvider { name, manifest: ModelProviderManifest }
```
I am unsure why do we need a concept of manifest at all. PUT and GET both can work with `ModelProvider`

Internally, the record can have a top level `name`

debajyoti-truefoundry:
```
OpenAIModelProvider
no_name

CustomModelProvider
name



input: OpenAIModelProvider | CustomModelProvider

ModelProviderRecord
name = input.type 
manifest

```

debajyoti-truefoundry (approved):

PR #88, opened 2026-08-06
Drop AgentDefinition.model; let ILLM own model identity

pranjaltrvd:
## Summary
- Remove the redundant `AgentDefinition.model` string; model identity now lives only on the bound `ILLM` client (`VercelAILLM` `providerConfig`).
- Change `ILLM` request types to `Omit<..., 'model'>` so callers no longer invent a model just to satisfy OpenAI's required field.
- Resolve dynamic sub-agent `agent_info.model` overrides into a real `modelClient` via `deps.llm`, instead of relabeling a stale client.

## Test plan
- [x] `pnpm --filter @truefoundry/utils-core typecheck`
- [x] `pnpm --filter @truefoundry/utils-core test`
- [ ] Smoke a turn with the default agent model
- [ ] Smoke a dynamic sub-agent that requests a different model and confirm the provider call uses that model's `modelId`

---

> [!NOTE]
> **Medium Risk**
> Breaking public shape on AgentDefinition/ILLM and changes which provider model sub-agents actually call; core turn path but localized to harness LLM wiring.
> 
> **Overview**
> **Removes the parallel `AgentDefinition.model` string** so catalog/wire model identity lives only on the bound `ILLM` (e.g. `VercelAILLM` `providerConfig`). `ILLM.create` / `createNonStream` now take `LLMCreateParams` types that **omit OpenAI’s required `model` field**; `AgentThread`, context compaction, and `VercelAILLM` stop setting or trusting request-body `model`, and stream chunk labels use the client’s catalog name instead.
> 
> **Fixes dynamic sub-agent model overrides:** `TurnResourceResolver.resolveAgentDefinition` picks `agent_info.model ?? spec.model.name` and calls `deps.llm(modelName)` for `modelClient`. `SessionHandle` no longer overlays `request.model` on a child definition while reusing a stale client.
> 
> Harness tests and public exports are updated for the slimmer `AgentDefinition` and new param types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1218b92283b9296eb9083c76fbdff4438b6a7605. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/harness/src/core/runtime/AgentDefinition.ts
debajyoti-truefoundry, line 17: `model, modelParams, modelClient`

Technically, `model` and `modelParams` are both embeddable within `modelClient`.

To make things consistent, my suggestion is to keep `model` and make `modelClient` independent of `model` or the model name for now.

Later, when we have the model set and whatnot, it will be a little bit easier.

model can be an argument for `create` in ILLM.

Please keep modelName, modelClient, and ModelParams clustered together in the `AgentDefinition`.

pranjaltrvd: Addressed in b5fe356:

- Kept `model` on `AgentDefinition` and clustered it with `modelClient` / `modelParams`.
- `model` is again the argument passed into `ILLM.create` (OpenAI params keep required `model`).
- `modelClient` stays independent of create-time model overrides for now (sub-agents may override `model` alone); embedding into the client can wait on model-set work.

debajyoti-truefoundry (approved):

PR #89, opened 2026-08-06
AGE-1578: Return 422 for semantic API errors

bhaveshpatel640:
AGE-1578: Return 422 for semantic API errors

---

| Surface | Route / source | Was | Now | Why |
|---|---|---|---|---|
| PUT MCP server (settings) | `PUT /api/v1/settings/mcp-servers` via `McpDcrConfigurationError` | 400 | 422 | Valid `auth.type: dcr` body, but remote cannot do DCR (e.g. no `registration_endpoint`). |
| Authorize MCP server | `GET …/mcp-servers/{name}/authorize` | 500 (DCR config) | 422 | Same permanent DCR misconfig; previously only 400 was forwarded, so 422 became 500. |
| `McpDcrConfigurationError` | error class constructor | 400 | 422 | Shared status carrier so handlers forwarding `statusCode` return 422. |
| PATCH session | `PATCH /api/v1/sessions/{id}` via `SessionStoreInvariantError` | 400 | 422 | Valid body, but named/ref sessions cannot replace their agent. |
| Create turn — sandbox required | `POST …/turns` / `agent_sandbox_required` | 400 | 422 | Valid turn input needs a sandbox (e.g. file upload) and none is configured. |
| Create turn — send vs thread state | `POST …/turns` / `invalid_send_input` | 400 | 422 | Valid input conflicts with thread state (unresolved tools/approvals, etc.). |
| Create turn — tool name collision | `POST …/turns` / `tool_name_collision` | 400 | 422 | Valid turn, but MCP tool names collide after sanitization. |
| ToolSet missing tools | `ToolSet` → `McpConnectionError` | 400 | 422 | Requested tools are well-named but not present on the MCP server. |

---

> [!NOTE]
> **Medium Risk**
> Breaking change for API/SDK clients that treated semantic harness and MCP errors as 400; status codes and error class names shift across sessions, turns, and MCP authorize/upsert.
> 
> **Overview**
> **HTTP semantics** are tightened so syntactic problems stay **400** while well-formed requests that cannot be satisfied return **422**, and upstream MCP OAuth/DCR failures return **424**.
> 
> MCP: permanent DCR misconfiguration (`McpDcrConfigurationError`, missing `registration_endpoint`) moves from **400** to **422** on settings upsert and authorize; transient DCR registration / authorization startup failures use **424** instead of **400**. Authorize now forwards **422** and **424** from `McpConnectionError` (fixing cases that previously became **500**). Missing enabled MCP tools in `ToolSet.listTools` now surface as **422**.
> 
> Sessions and turns: named-session agent update invariant failures return **422** instead of **400**. On **POST …/turns**, harness rejections are mapped by error code—`invalid_file_input` stays **400**; `invalid_send_input`, `agent_sandbox_required`, and `tool_name_collision` become **422**; `capability_state_error` remains an internal **500**.
> 
> Harness rename: `invalid_agent_input` / `InvalidAgentInputError` becomes **`invalid_file_input`** / **`InvalidFileInputError`** for file URI validation. OpenAPI, route docs, regenerated SDK clients (`UnprocessableEntityError`, `FailedDependencyError`), and tests are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3409af5a5666623e711e02afdba40f5cf39a1e33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/server/src/apis/turns.ts
```diff
@@ -483,7 +483,17 @@ export function createTurnsRouter(deps: TurnsRouterDeps) {
         return c.json({ error: { message: error.message } }, 404);
       }
       if (error instanceof AgentHarnessError && !(error instanceof McpConnectionError)) {
-        return c.json({ error: { message: error.message } }, 400);
+        switch (error.code) {
+          case 'invalid_agent_input':
+            return c.json({ error: { message: error.message } }, 400);
```
debajyoti-truefoundry, line 488: 422

debajyoti-truefoundry (approved):

packages/server/src/apis/turns.ts
bhaveshpatel640: i misread this against `invalid_send_input`


So 
- `invalid_send_input` is raised when input is invalid against the turn's current state, like `User message sent while approvals/questions pending` or `User messages while sub-agents running` and so on (Updated to 422)
- `invalid_agent_input` is raised only in one place, when a user message's file part is malformed - unparseable data or empty base64 payload (Keeping 400 only)

I think this cases should return 400 only, since file content is invalid here?


Still checking other such cases for 400 to 422...

bhaveshpatel640: ### MCP authorize: should upstream DCR failures stay 400?

These look like **upstream authorization-server faults**, not bad client input. Verifying if it's a valid candidate for 502 also.

| Case | File | Today | Proposal |
|---|---|---:|---|
| `registerClient` fails | `mcpDcr.ts` (~56–60) | 400 | 502/422 |
| Retry `registerClient` fails | `mcpDcr.ts` (~65–69) | 400 | 502/422 |
| Registration response missing `client_id` | `mcpDcr.ts` (~73–76) | 400 | 502/422 |
| `startAuthorization` throws | `mcpDcr.ts` (~179) | 400 | 502/422 |


@heerambavi1998 / @thesujai 
Please review these cases once, here we are returning blanket 400

heerambavi1998: upstream shouldn't be 502, these are not errors on our side. Can we do 422? 
Also, specifically for the error where it returns DCR not supported, it can be 422 ( we are trying DCR on non DCR mcp)

bhaveshpatel640: > upstream shouldn't be 502, these are not errors on our side. Can we do 422?
 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/502
- The HTTP 502 Bad Gateway status code indicates that a server was acting as a gateway or proxy and that it received an invalid response from the upstream server.

I think we should be fine with 422 for now, we can move to 502 later if needed, 502 can also be misread as our server being down.


---
> Also, specifically for the error where it returns DCR not supported, it can be 422 ( we are trying DCR on non DCR mcp)

DCR not supported is already updated to 422 (McpDcrConfigurationError when registration_endpoint / auth+token endpoints are missing).

bhaveshpatel640: We will have 424 in these cases, had a word with Deb

bhaveshpatel640: `invalid_agent_input` seems broad, we can fix this naming here

bhaveshpatel640: Renamed to invalid_file_input

PR #218, opened 2026-08-13
Revert "Revert "feat: updated capabilities and save modal flow""

chiragjn:
Reverts truefoundry/trueforge#217


chiragjn:
This cannot be merged because it has manual change to `sdk`

PR #238, opened 2026-08-13
fix: move DCR remote I/O under txn

heerambavi1998:
Cleanup multiple reads in and out of txn as well

---

> [!NOTE]
> **Medium Risk**
> Changes MCP OAuth persistence and runs remote DCR HTTP inside a DB transaction on PUT (bounded by 15s OAuth timeout vs 30s idle-in-txn), which affects auth correctness and connection hold time but is explicitly documented.
> 
> **Overview**
> **MCP server PUT (DCR)** now runs dynamic client registration inside the same `withTransaction` as the row lock, header secret merge, upsert, and `saveClient`, so a failed AS registration rolls back the server write instead of leaving a half-updated row. Create still performs DCR **before** the txn. The pre-transaction `getServer` read for PUT is removed; `getServerForUpdate` inside the txn is the single source of truth. Failed DCR on upsert maps to **422** via `McpConnectionError`.
> 
> **OAuth flow** is tightened: `buildMcpAuthorizationUrl` expects an already-resolved `client` (registration is centralized in `resolveMcpAuth` via `ensureMcpClientRegistered` before refresh or a new auth URL). The public OAuth callback router no longer needs `withTransaction`; on token exchange `invalid_client`, only the shared OAuth **client** is deleted (per-user tokens are not cleared in that path).
> 
> Minor API cleanup: settings MCP routers reuse `McpServersRouterDeps`, and the disconnect route is renamed to `deleteAuthorizeMcpServerRoute` / `delete_authorize`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20a76643bef8fc008706a01593390216a5c8ed60. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/server/src/apis/mcpServers.ts
```diff
@@ -210,8 +183,14 @@ export function createSettingsMcpServersRouter<TTransaction>(deps: SettingsMcpSe
...
-        if (dcrClientToSave !== undefined) {
-          await deps.mcpServerStore.saveClient({ id: saved.id, record: dcrClientToSave }, transaction);
+        if (manifest.auth?.type === 'dcr') {
+          const dcrClient = await createMcpOAuthClient({
+            mcpServerUrl: manifest.url,
+            mcpServerName: manifest.name,
+            redirectUri: mcpOAuthCallbackUrl(),
+            clientName: configuration.MCP_DCR_OAUTH_CLIENT_NAME,
+          });
+          await deps.mcpServerStore.saveClient({ id: saved.id, record: dcrClient }, transaction);
```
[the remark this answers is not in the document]
heerambavi1998: Yes this is known caveat.

packages/server/src/apis/mcpServers.ts
```diff
@@ -276,6 +227,26 @@ export function createSettingsMcpServersRouter<TTransaction>(deps: SettingsMcpSe
           incoming: incomingManifest,
           existing: existing?.manifest,
         });
+
+        let dcrClientToSave: OAuthClientRecord | undefined;
+        if (manifest.auth?.type === 'dcr') {
+          const existingClient =
+            existing !== undefined ? await deps.mcpServerStore.getClient({ id: existing.id }, transaction) : undefined;
```
sr07asthana:
```suggestion
            existing ? await deps.mcpServerStore.getClient({ id: existing.id }, transaction) : undefined;
```

packages/server/src/apis/mcpServers.ts
```diff
@@ -276,6 +227,26 @@ export function createSettingsMcpServersRouter<TTransaction>(deps: SettingsMcpSe
...
+          // Register when: brand-new server, client was cleared (e.g. invalid_client), or MCP URL changed
+          // (resource/AS may differ; reuse would keep stale oauth_server/client for the old AS).
+          // TODO: make manifest URL immutable after create so re-DCR / stale OAuth tokens are not possible via PUT.
+          const urlChanged = existing !== undefined && existing.manifest.url !== manifest.url;
```
sr07asthana:
```suggestion
          const urlChanged = existing && existing.manifest.url !== manifest.url;
```

packages/server/src/mcp/auth/mcpDcr.ts
```diff
@@ -171,44 +171,49 @@ export async function createMcpOAuthClient(params: {
   });
 }
 
-export async function ensureMcpClientRegistered(params: {
-  mcpServerStore: IOAuthClientStore;
+export async function ensureMcpClientRegistered<TTransaction = never>(params: {
```
sr07asthana: just for my understanding what is the reason we need to type as TTransaction = never?

sr07asthana (approved):

packages/server/src/apis/mcpServers.ts
sr07asthana: do we need check for undefined? i see at a lot of places in this PR, 
a simple `existing? await deps.mcpServerStore.getClient({ id: existing.id }, transaction) : undefined;` should work.

packages/server/src/mcp/auth/mcpDcr.ts
heerambavi1998: its not actually needed, was for non db implementations

PR #294, opened 2026-08-17
Cancellation fails after a restart in distributed mode

heerambavi1998:
## Summary

Cancellation fails after a restart in distributed mode. Fix this by freezing the turn in such scenarios
Closes https://linear.app/truefoundry/issue/AGE-1842/after-a-restart-in-postgres-mode

---

> [!NOTE]
> **Medium Risk**
> Changes distributed session cancellation and durable turn state; unconfirmed freezes may race with a still-running owning executor, though store fencing limits impact.
> 
> **Overview**
> Fixes sessions stuck in `running` after executor restarts or failed distributed cancel (e.g. Postgres/distributed mode).
> 
> **Session cancel** no longer fails with 412/424 when the owning executor is gone or Redis cannot confirm abort. `cancelSessionTurn` now **freezes** the turn in the store via `SessionHandle.freezeTurn` so the session can start a new turn. Confirmed local or peer abort (HTTP 200) still relies on the executor path to persist terminal state; unconfirmed cases log a warning and freeze anyway (the owning replica may still be running; first-terminal-write-wins applies).
> 
> **Store API:** `freezeAndGetTurn` takes a caller-supplied `CancellationReason` (barge-in `cancelled-for-next-turn` vs explicit `client-cancelled`). `SessionHandle.freezeTurn` wraps this for cancel and `createTurn` previous-turn cancellation.
> 
> Tests cover freeze-on-missing-local-run, Redis timeout/unreachable, and peer 412.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2729c095e53888865e4919995a7b88ee9d499995. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/apis/sessions.ts
```diff
@@ -112,17 +117,57 @@ export function cancelSessionTurnPeerHandler(activeTurns: ActiveTurnRegistry): R
...
 export interface CancelTurnDeps {
   activeTurns: ActiveTurnRegistry;
-  sessionStore: Pick<ISessionStore, 'getTurn'>;
+  sessionStore: Pick<ISessionStore, 'getTurn' | 'freezeAndGetTurn'>;
   redis?: RedisClientType | undefined;
+  logger: Pick<Logger, 'warn'>;
+}
+
+/** Freeze a running turn; missing turns are a no-op (already gone). */
+async function freezeTurn(
```
debajyoti-truefoundry: Does it make sense to add it in the session or turn handle?

debajyoti-truefoundry: I suggest you take a look at the turn handle once please.

debajyoti-truefoundry (approved):

packages/trueforge/src/apis/sessions.ts
heerambavi1998: cancel might not even have turn handle, because the process itself is gone. 
Keeping it session level makes sense, as session needs to handle it when turn is gone. Added there.

PR #409, opened 2026-08-22
feat(sandbox): stream file downloads end to end

### LINKED ISSUE #283: Stream sandbox file downloads end to end
## Problem

Sandbox file downloads currently buffer the complete file in each provider and then copy that buffer again into the HTTP response. The route contains a TODO to stream the response, but the current `SandboxProvider.downloadFile` contract and all implementations return `Promise<Buffer>`.

A route-only `ReadableStream` wrapper would not reduce peak memory because the provider has already materialized the complete file.

## Proposed direction

Introduce a genuinely incremental provider contract, then stream it through the turns download route while preserving the current ownership/path checks, maximum file-size enforcement, response headers, and error mapping.

Likely surfaces:

- `SandboxProvider` file-download contract
- Daytona provider
- TFY sandbox transport
- local sandbox provider
- turns download route
- provider contract and route tests

## Design questions

- Should the provider return a Web `ReadableStream<Uint8Array>`, an async iterable, or a response object containing stream plus known size?
- Does the Daytona SDK expose a streaming file API, or is a lower-level transport required?
- What chunked endpoint/protocol should the TFY sandbox use instead of base64 embedded in an exec result?
- Is a temporary buffered compatibility fallback acceptable, or should the contract change only when every provider streams genuinely?

## Acceptance criteria

- Download bytes begin reaching the client before the complete file is loaded.
- Peak application memory is bounded by chunk/backpressure behavior rather than file size.
- Existing maximum-size and tenant/path controls remain enforced.
- Client cancellation stops provider reads.
- Provider contract tests verify progressive delivery and error behavior.

aniruddhaadak80:
## Summary

Sandbox file downloads currently buffer the complete file inside every provider and then copy that buffer again into the HTTP response. This PR changes `SandboxProvider.downloadFile` to return `{ size, stream }` and streams bytes end to end, so peak memory is bounded by chunk/backpressure rather than file size, the client starts receiving bytes before the whole file is read, and disconnects stop provider reads.

Design notes for the questions raised in the issue:

- **Contract shape**: a plain object `{ size: number; stream: ReadableStream<Uint8Array> }`. The size comes from the existing pre-download stat (which also keeps the is-dir and max-size checks ahead of any bytes), so the route still sends an exact `Content-Length`; errors raised before the first byte stay mappable to HTTP statuses, while mid-stream failures surface as an errored stream detectable against Content-Length.
- **Cancellation**: the route wires `req.raw.signal` into the provider call, consumer cancellation propagates through the stream's `cancel()`, and both stop provider reads (process-group kill locally, stream destroy on Daytona, no further execs on TFY).
- **TFY transport**: the sandbox server has no file endpoint today, so instead of base64-buffering the whole file in one exec, bytes are pulled as sequential fixed-window execs (`tail -c +N | head -c M | base64 -w0`). Memory stays bounded per 2 MiB chunk with no server-side changes; a temporary fully-buffered fallback was deliberately avoided.

Closes #283

## Changes

- `trueforge-core`: `SandboxProvider.downloadFile` now returns `SandboxFileDownload` (`size` + single-consumption web stream) and accepts an optional abort `signal`.
- `trueforge-core`: new shared `boundedFileStream` helper wraps incremental chunks in one canonical implementation that enforces the download cap *while* streaming (guards against a file growing after stat) and returns generator cleanup on cancel/error.
- **Local provider**: streams raw stdout Buffers via a new incremental SRT session runner (`startSupervisorStdoutStream`, binary-safe — the old path utf8-decoded buffered output, which forced base64). Spawn/wrap failures still throw before any byte moves so the route can answer with a proper status; `/bin/cat` runs under the same SRT wrap policy as exec/upload.
- **Daytona provider**: uses `sandbox.fs.downloadFileStream(...)` (already exposed by `@daytona/sdk`) adapted into the contract stream, keeping stat-first checks and the 404 mapping.
- **TFY provider**: windowed exec pulls as above, with per-window length verification (detects files mutated mid-download), first-window failure mapped to `SandboxFileNotFoundError`, and caller-abort support combined with the exec timeout via `AbortSignal.any`.
- Turns download route: sends the stream as the body with unchanged headers (`Content-Type`, exact `Content-Length` from the stat size, `Content-Disposition`, `Cache-Control`) and unchanged error mapping; removed the now-dead buffering helper.
- Added `.changeset/stream-sandbox-file-downloads.md` (published packages changed). No generated OpenAPI/SDK edits — wire shape is unchanged.

## How was this tested?

- New unit tests:
  - `boundedFileStream.test.ts`: ordered forwarding, cap enforcement (offending chunk withheld), source-error propagation, cleanup on consumer cancel.
  - `daytonaDownload.test.ts`: streamed round-trip + stat size, dir/oversize rejected before opening the stream, missing-file mapping, streaming-cap re-enforcement when a file grows after stat.
  - `tfyDownload.test.ts`: multi-window reassembly across the 2 MiB boundary (command shape asserted), empty file, pre-byte rejections, short-read failure, first-window failure mapping.
- Provider contract suite now verifies incremental round-trip of a 1 MiB payload (multi-chunk), plus missing-file and directory error contracts for every backend.
- Route tests (`sandboxFileDownload.test.ts`): progressive delivery through the Hono response (first chunk observed while the producer is parked), exact headers, `signal` passed down, and SandboxError → status mapping.
- Updated the local sandbox smoke test to consume the new stream API.
- Locally: `pnpm --filter @truefoundry/trueforge-core typecheck/test`, `pnpm --filter @truefoundry/trueforge typecheck/test`, targeted jest suites, eslint and prettier on all touched files pass. (Windows dev box: repo-wide `pnpm build`/`format:check` are blocked by Unix-only scripts/CRLF checkout and two local-sandbox suites fail on POSIX-mode assertions — all verified identical on a clean `main` stash, unrelated to this change.)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes the sandbox download contract and HTTP body path (streaming, abort, size-cap re-check) across all providers. Not auth, but a user-facing data transfer path where mid-stream failures become truncated bodies.
> 
> **Overview**
> Sandbox file downloads no longer buffer the whole file. `SandboxProvider.downloadFile` now returns `{ size, stream }` plus an optional abort `signal`, so peak memory tracks chunk size, `Content-Length` still comes from the pre-stat, and client disconnects stop provider reads.
> 
> Shared `boundedFileStream` wraps incremental chunks, re-enforces the download cap while streaming (withholds past-cap bytes), and runs generator cleanup on cancel. Daytona uses `downloadFileStream`; TFY pulls 2 MiB windows via sequential execs instead of one base64 dump; local SRT cats raw stdout through a new `startSupervisorStdoutStream`.
> 
> The turns download route pipes the stream as the body (same headers) and maps pre-byte `SandboxError`s to JSON statuses as before. Mid-stream failures surface as a truncated body vs `Content-Length`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78d534b0763c09eea301aeb0bf8a84dd4db7ac84. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/stream-sandbox-file-downloads.md
+++ b/.changeset/stream-sandbox-file-downloads.md
@@ -1,0 +1,6 @@
+---
+"@truefoundry/trueforge-core": patch
+"@truefoundry/trueforge": patch
+---
+
+Stream sandbox file downloads end to end. `SandboxProvider.downloadFile` now returns `{ size, stream }` so the download route can send bytes incrementally (bounded by chunk size, not file size), keep the exact `Content-Length`, stop provider reads when the client disconnects, and re-enforce the size cap while streaming.
--- a/packages/trueforge-core/src/core/index.ts
+++ b/packages/trueforge-core/src/core/index.ts
@@ -142,6 +142,7 @@ export type { CodeModeLogger } from './sandbox/codeMode/CodeModeDispatcher';
 export type { CodeModeClientInstall, CodeModeTransport } from './sandbox/codeMode/CodeModeTransport';
 export { CodeModeErrorSourceSchema, CodeModeReplySchema, CodeModeRequestSchema } from './sandbox/codeMode/types';
 export type { CodeModeErrorSource, CodeModeReply, CodeModeRequest } from './sandbox/codeMode/types';
+export { boundedFileStream } from './sandbox/provider/boundedFileStream';
 export { DaytonaSandboxProvider } from './sandbox/provider/DaytonaProvider';
 export type { DaytonaSandboxProviderOptions } from './sandbox/provider/DaytonaProvider';
 export { absolutizeRelativeExecEnv } from './sandbox/provider/execEnv';
@@ -154,6 +155,7 @@ export type {
   SandboxBuildMetadata,
   SandboxBuildStatus,
   SandboxExecParams,
+  SandboxFileDownload,
   SandboxInit,
   SandboxProvider,
 } from './sandbox/provider/Provider';
--- a/packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
+++ b/packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
@@ -6,6 +6,9 @@ import { randomUUID } from 'node:crypto';
 import { join } from 'node:path/posix';
 import type { Logger } from 'winston';
 import { extractErrorLogFields } from '../../util/errorLogFields';
+import type { CodeModeTransport } from '../codeMode/CodeModeTransport';
+import { CodeModeNatsTransport } from '../codeMode/nats/CodeModeNatsTransport';
+import { DEFAULT_PREVIEW_URL_EXPIRY_SECONDS, DEFAULT_SANDBOX_NATS_WS_PORT } from '../constants';
 import {
   SandboxFileNotFoundError,
   SandboxFileTooLargeError,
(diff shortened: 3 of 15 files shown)


aniruddhaadak80:
﻿Heads up for reviewers: the CI / Generate SDK workflow runs are sitting in `action_required` because this is my first contribution here — they need a one-click 'Approve and run workflows' to start. Everything those jobs run was verified locally (typecheck, unit suites incl. the new ones, eslint/prettier on touched files); details in the PR description.

PR #434, opened 2026-08-25
Fixes #420: Harden Helm chart security defaults

### LINKED ISSUE #420: Helm chart ships no default securityContext — server container runs as root with full capabilities
### What problem are you trying to solve?

The Helm chart deploys the TrueForge server with **no pod/container security context at all**, so on any default Kubernetes cluster the container runs as **root** with all capabilities:

```yaml
# charts/trueforge/values.yaml (main @ 506bf5c)
podSecurityContext: {}
securityContext: {}
```

```yaml
# charts/trueforge/templates/deployment.yaml — both blocks render only when values are provided
{{- with .Values.podSecurityContext }}
securityContext:
  {{- toYaml . | nindent 12 }}
{{- end }}
...
{{- with .Values.securityContext }}
securityContext:
  {{- toYaml . | nindent 12 }}
{{- end }}
```

Neither `Dockerfile` nor `Dockerfile.dev` sets a `USER`, which confirms the image itself has no non-root runtime user either. The result on stock values: a network-facing server process (API + UI, plus sandbox orchestration) running as UID 0, with full capabilities and a writable root filesystem.

For a workload whose threat model includes agent-driven code execution in sandboxes, "container escape mitigations cost nothing" applies strongly here: an attacker who gets code execution inside the server process currently lands as root in the container.

### Proposed solution

Ship secure-by-default (overridable) contexts, e.g.:

```yaml
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 10001
  runAsGroup: 10001
  fsGroup: 10001
  seccompProfile: { type: RuntimeDefault }
securityContext:
  allowPrivilegeEscalation: false
  readOnlyRootFilesystem: true
  capabilities: { drop: ["ALL"] }
```

Concretely this needs:

1. A non-root `USER` in both Dockerfiles (and chown of any paths the server writes).
2. An empty-dir mount + writable paths for anything the server mutates at runtime (SQLite is not used in the Helm topology since `STANDALONE=false`, so this should mostly be tmp/logs).
3. Chart CI (`pnpm chart:lint` / template tests) extended to assert the rendered Deployment carries a security context.

Users who genuinely need root can still override via the existing values.

### Alternatives considered

- Documenting recommended values in the README instead of changing defaults — defaults are what actually ships; PSA `restricted` namespaces (increasingly common) will reject these pods outright today.
- PSA labels only — orthogonal; per-workload contexts are still needed for `restricted`.

### Additional context

Related hardening observations in the same chart (happy to split if preferred): no `NetworkPolicy` templates, and no resource requests/limits by default. The core ask here is the root-by-default runtime, which is the highest-impact item.

rohanmalhotracodes:
## Summary

Hardens the TrueForge Helm deployment so the server container no longer runs as root by default. The chart now renders restricted pod/container security contexts and mounts `/tmp` as writable storage so `readOnlyRootFilesystem: true` works.

Closes #420

## Changes

- Added non-root runtime user `10001:10001` in `Dockerfile` and `Dockerfile.dev`
- Added secure default `podSecurityContext` and container `securityContext`
- Added default `/tmp` `emptyDir` volume mount for writable temp storage
- Added chart render coverage for the Deployment security context
- Added Helm chart CI checks for chart changes
- Updated chart README with the new security defaults

## How was this tested?

- `pnpm build`
- `pnpm test`
- `pnpm typecheck`
- `pnpm lint:ci`
- `pnpm chart:deps`
- `pnpm chart:lint`
- `pnpm chart:template`
- `pnpm test:chart-security`

Before 
main-before

After
helm-after

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes default pod runtime identity, filesystem writability, and resource limits for Helm installs; upgrades may need values overrides if workloads relied on root or empty resource defaults.
> 
> **Overview**
> Hardens TrueForge Kubernetes deployments by aligning container images and the Helm chart with **restricted security defaults** (closes #420).
> 
> **Docker images** (`Dockerfile`, `Dockerfile.dev`) now create user/group **10001** and run the server process as **non-root** instead of root.
> 
> **Helm chart** defaults add pod-level `runAsNonRoot` / UID **10001**, container `readOnlyRootFilesystem`, dropped capabilities, and concrete **CPU/memory/ephemeral-storage** requests and limits. The Deployment always mounts a writable **`/tmp` `emptyDir`** (optionally sized from `resources.limits.ephemeral-storage`) so read-only root filesystem still works.
> 
> **CI** path-filters chart-related changes and runs a gated **`chart-checks`** job (`pnpm chart:deps`, `pnpm chart:lint`); the aggregate **`check`** job treats skipped/success chart checks like other optional suites. Chart **README** documents the new defaults and `/tmp` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b58347d0fbe00545399ea11eaa330e6142aca988. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/scripts/chart-security-context.test.sh
chiragjn, line 1: We don't need to test this

rohanmalhotracodes: Removed the custom chart security-context test and wiring.

charts/trueforge/templates/deployment.yaml
```diff
@@ -75,17 +75,21 @@ spec:
...
             {{- toYaml . | nindent 12 }}
-          {{- end }}
+            {{- end }}
         {{- with .Values.extraContainers }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.extraVolumes }}
       volumes:
+        - name: tmp
+          emptyDir: {}
```
chiragjn, line 95: 1. let's have default values for in `resources` in values.yaml

```
resources:
  requests:
    cpu: 100m
    memory: 256Mi
    ephemeral-storage: 256Mi
  limits:
    cpu: 200m
    memory: 512Mi
    ephemeral-storage: 512Mi
```

and let's make this empty dir size equal to resources.limits.ephemeral-storage if it is not undefined

Dockerfile
```diff
@@ -14,7 +14,8 @@ WORKDIR /app
 # HOST=0.0.0.0 so Kubernetes Service/probe traffic reaches the process.
 ENV NODE_ENV=production \
     STANDALONE=false \
-    HOST=0.0.0.0
+    HOST=0.0.0.0 \
+    HOME=/tmp
```
chiragjn: ideally I don't want to change HOME,
rather the code should take care to always use platform specific tmp dir to write to

Let's look at all code and see if something is writing outside of /tmp in non-standalone mode

Dockerfile
```diff
@@ -23,7 +24,11 @@ RUN test -n "$APP_VERSION" || (echo "APP_VERSION build-arg is required" >&2 && e
 RUN npm install --omit=dev "@truefoundry/trueforge@${APP_VERSION}" \
   && npm cache clean --force
 
+RUN groupadd --gid 10001 trueforge \
+  && useradd --uid 10001 --gid trueforge --home-dir /tmp --no-create-home --shell /usr/sbin/nologin trueforge
```
chiragjn: related to above comment

```suggestion
  && useradd --uid 10001 --gid trueforge --shell /usr/sbin/nologin trueforge
```

Dockerfile.dev
```diff
@@ -91,7 +92,11 @@ COPY --from=frontend-builder /app/packages/frontend/dist ./packages/trueforge/di
 
 WORKDIR /app/packages/trueforge
 
+RUN groupadd --gid 10001 trueforge \
+  && useradd --uid 10001 --gid trueforge --home-dir /tmp --no-create-home --shell /usr/sbin/nologin trueforge
```
chiragjn:
```suggestion
  && useradd --uid 10001 --gid trueforge trueforge
```

this is a dev image, I would like to be able to shell in for debugging

Dockerfile.dev
```diff
@@ -71,7 +71,8 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
 # runner: minimal image with prod node_modules + built artifacts.
 # ---------------------------------------------------------------------------
 FROM base AS runner
-ENV NODE_ENV=production
+ENV NODE_ENV=production \
+    HOME=/tmp
```
chiragjn: same point as https://github.com/truefoundry/trueforge/pull/434/changes#r3860523039

Dockerfile
chiragjn: I did some analysis with Cursor - we should be good to go here
We don't write to rootfs in `STANDALONE=false`

charts/trueforge/templates/deployment.yaml
rohanmalhotracodes: Added the default `resources` values in `values.yaml`. 



The `/tmp` `emptyDir` now uses `resources.limits.ephemeral-storage` as `sizeLimit` when it is set, and falls back to plain `emptyDir: {}` when it is not.

Dockerfile
rohanmalhotracodes: Applied useradd` change: removed the `/tmp` home override and kept the `nologin` shell.

Dockerfile.dev
rohanmalhotracodes: Applied suggestion and changed `useradd` to keep shell-in debugging possible.

Dockerfile.dev
rohanmalhotracodes: Removed `HOME=/tmp` from `Dockerfile.dev` as well.

Dockerfile
rohanmalhotracodes: Addressed the points mentioned.

chiragjn (approved):

Issue #427, opened 2026-08-25
`npx @truefoundry/trueforge` command fails to start the server on Windows Terminal (PS, CMD, and Git Bash)


DracFiendMG:
### What happened?

When executing the command `npx @truefoundry/trueforge` on Windows Terminal (PS, CMD, and Git Bash), its unable to start the server.

Screenshot:
<img width="1246" height="374" alt="Image" src="https://github.com/user-attachments/assets/374d7ba0-becc-4042-b5c7-e23ccea19c07" />

### Steps to reproduce

1. Run `npx @truefoundry/trueforge` on anyone of the Windows Terminal (PS, CMD, and Git Bash)
2. Press 'y' to proceed with the installation
3. Check the logs after installation
4. The logs indicate that the server has failed to start.

### How are you running TrueForge?

npx (standalone / SQLite)

### Version

0.1.4

### Relevant logs or screenshots

```shell
11:16:34 warn Local sandbox fallback is unavailable {"reason":"LocalSandboxProvider supports macOS and Linux only (got win32)"}
Failed to start server: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

### Environment

OS: Windows 11
Node: v24.15.0
Terminal: Powershell, Command Prompt, Git Bash

debajyoti-truefoundry:
@DracFiendMG Thanks for reporting this.

Would you be interested in picking this up and opening a PR?

DracFiendMG:
> [@DracFiendMG](https://github.com/DracFiendMG) Thanks for reporting this.
> 
> Would you be interested in picking this up and opening a PR?

Sure, please assign it to me, I'll look into it

jayesh9747:
<img width="1026" height="655" alt="Image" src="https://github.com/user-attachments/assets/ec4dcd70-29a5-4169-a41d-821a61ad5f0a" />

its working for me , did u install the docker ?? i think its use the docker for image creation and installtion

DracFiendMG:
[...]

PR #374, opened 2026-08-20
Feat: add cross platform support

SidakSethi-Singh:
Fixes #427 

Summary

This PR resolves compatibility issues when building, linting, and running the test suite on Windows hosts.

Changes made
Clean scripts — replaced Unix-specific rm -rf build-clean commands with a cross-platform node -e "fs.rmSync(...)" equivalent.
Environment variables — updated inline env var syntax in package scripts (via cross-env) so they work in both Windows Command Prompt/PowerShell and Unix shells.
Sandbox paths — updated layout path handling in Sandbox.ts to use node:path/posix instead of node:path, since the sandbox runtime is POSIX-based regardless of host OS and needs consistent forward-slash paths.
Smoke test / platform checks — updated packageExports.smoke.mjs to run correctly under pnpm.cmd on Windows.
UDS transport — CodeModeUdsTransport.ts skips owner-only permission checks and Unix socket path-length validation on Windows, since Windows named pipes don't have the same constraints as Unix domain sockets, and uses Windows junctions for temp test directories instead of symlinks.
Verification
npx pnpm build compiles successfully
npx pnpm lint:ci passes without errors
npx pnpm test — all 374 unit tests pass
npx pnpm test:store:sqlite — all 56 tests pass

All changes were tested locally on a Windows host.

---

> [!NOTE]
> **Medium Risk**
> Changes how migrations are dynamically imported and gates releases on a new Windows smoke job; scoped to Windows compatibility but touches DB startup and the publish pipeline.
> 
> **Overview**
> Improves **native Windows** support for the published `npx @truefoundry/trueforge` flow while keeping **source development** documented as Unix/WSL-only.
> 
> **Runtime fixes:** Kysely SQLite/Postgres migrations now load compiled `.js` files via `pathToFileURL` (`importAbsoluteModule`), fixing dynamic import on Windows. Sandbox guest path layout in `Sandbox.ts` uses **`node:path/posix`** so remote paths stay forward-slash regardless of host OS.
> 
> **Build/scripts:** Unix-only `rm -rf` / `chmod +x` in package scripts are replaced with **`rm-path.mjs`** / **`chmod-path.mjs`**, and **`cross-env`** sets `NODE_ENV`, `STANDALONE`, and `NODE_OPTIONS` in npm scripts on Windows.
> 
[...]

PR #463, opened 2026-08-27
fix(trueforge): match the redaction sentinel exactly and publish less of a secret

### LINKED ISSUE #423: Secret redaction: sentinel matched by substring (real keys containing ***REDACTED*** silently dropped) and 3+3 char prefix/suffix leak for short secrets
### What happened?

`packages/trueforge/src/utils/secretRedaction.ts` — the module guarding every settings secret (model provider API keys, Daytona key, MCP static headers) — has two concrete correctness/disclosure defects:

```ts
// main @ 506bf5c
export const SECRET_REDACTION = '***REDACTED***';
const MIN_LENGTH_FOR_PREFIX_SUFFIX = 10;

/** Response mask: prefix/suffix for longer secrets; full mask when too short to hide. */
export function toRedactedSecretValue(secret: string): string {
  if (secret.length < MIN_LENGTH_FOR_PREFIX_SUFFIX) {
    return SECRET_REDACTION;
  }
  return `${secret.slice(0, 3)}-${SECRET_REDACTION}-${secret.slice(-3)}`;
}

/**
 * True when the client sent a redacted stand-in from a prior GET.
 */
export function isRedactedSecretValue(value: string): boolean {
  return value.includes(SECRET_REDACTION);
}
```

**Defect 1 — sentinel matched by substring, so legitimate values can be silently discarded.**
`isRedactedSecretValue` uses `.includes()`. Any *real* secret containing the literal string `***REDACTED***` (e.g. a header value like `Bearer token=***REDACTED***-demo` used against a mock server, or an API key generated with unusual charset policies) is classified as "keep stored" by `resolveStoredSecretValue`, which then either silently persists the OLD secret or throws a confusing 400 "API key is required" (`apis/sandboxProviders.ts`). The user's newly typed value is dropped without any error explaining why. Sentinel matching must be exact (`value === SECRET_REDACTION`) — the redaction function only ever emits that exact standalone string.

**Defect 2 — prefix/suffix leak exposes most of short secrets.**
For any secret ≥10 chars the response reveals the first **and** last 3 characters: a 10-char secret shows 6 of its 10 characters (`abc***REDACTED***xyz`); a 12-char key leaves only 6 unknown. Combined with typical API-key namespaces (e.g. keys starting with fixed prefixes), this meaningfully shrinks search space and confirms guesses. Every other major provider shows at most a fixed-length suffix (usually last 4). Either raise `MIN_LENGTH_FOR_PREFIX_SUFFIX` substantially (~24+) or show suffix-only, e.g. `…last4`.

**Defect 2b (same file):** because the masked form is itself ≥10 chars, a round-trip GET→PUT of an unchanged form re-sends the mask; that path is handled — but a GET followed by a PUT where the user edits *other* fields relies on defect 1's substring check to keep secrets, which is exactly the fragile coupling above.

### Steps to reproduce

Defect 1:
1. Open Settings → Model providers; create one with api_key `sk-***REDACTED***-test` (≥10 chars, contains the sentinel).
2. Save → then GET: response shows the stored value was NOT what was typed (resolveStoredSecretValue treated it as keep) — on create it 400s with "API key is required" instead.

Defect 2:
1. Create a model provider with a 12-character api_key.
2. `GET /api/v1/settings/model-providers` returns `abc***REDACTED***xyz` — 6 real key characters exposed to every admin viewer/screen-share/log capture of the response.

### Expected behavior

- Exact-match sentinel comparison.
- Masking policy that never reveals more than a bounded suffix (e.g. last 4) and only for long-enough secrets; behavior is currently pinned via `tests/unit/apis/modelProviders.test.ts` and `tests/unit/apis/sandboxProviders.test.ts`, which should be extended with dedicated cases for both defects.

### How are you running TrueForge?

npx (standalone / SQLite)

### Version

main @ 506bf5c

### Environment

OS: any
Node.js version: 22.14+

ethanstoner:
Fixes #423.

## Both defects, and one correction to the proposed fix

**Defect 1 — sentinel matched by substring.** `isRedactedSecretValue` used `.includes()`, so any real secret containing `***REDACTED***` anywhere was read as "keep the stored one". The value the user typed was discarded with nothing saying so: the old secret survived an update, and a create failed with a message about a missing key. It is now a whole-value test.

**One correction.** The issue proposes `value === SECRET_REDACTION`, on the grounds that "the redaction function only ever emits that exact standalone string". It does not — that is only the short-secret branch. A secret at or above the threshold masks to the sentinel *plus* a suffix, and a client round-tripping an unedited form sends that back. Exact-equality-only would classify it as a real value and **store the mask as the secret**, which is a worse bug than the one being fixed. So the test matches the two shapes `toRedactedSecretValue` actually emits, anchored to the whole value, and nothing else.

A secret whose entire content is exactly one of those two shapes remains indistinguishable from a mask. Nothing in-band can separate them — only a separate `keep` field beside the value could, which is an API change rather than a fix. Left alone, and documented in the module so the next reader does not have to rediscover it.

**Defect 2 — prefix/suffix leak.** Any secret of ten characters or more came back as first three + last three, so a ten-character key published six of its ten to every viewer of a settings response, screen share, or captured log. Now: last four only, and only for secrets of at least twelve, so eight characters stay unknown at the threshold. Shorter secrets are masked completely. Suffix-only matches what other providers publish, as the issue suggests.

```
before   toRedactedSecretValue('sk-ant-01234')  ->  sk-***REDACTED***-234
after    toRedactedSecretValue('sk-ant-01234')  ->  ***REDACTED***-1234
```

## Two existing tests asserted the defect

`mcpServers.test.ts` and `sandboxProviders.test.ts` each sent a redacted-*looking* value that was not the stored secret's mask (`oth-***REDACTED***-xxx`) and asserted the stored secret survived, with the comment:

> `// Any value containing ***REDACTED*** is treated as keep, not only the exact GET mask.`

That is defect 1 written down as intended behaviour, so both now assert the opposite: a value that merely contains the sentinel is stored like any other rotation. Flagging it explicitly since inverting an existing assertion deserves a look rather than a skim.

The MCP tests also hardcoded the wire form (`'Bea-***REDACTED***-ken'`). That is now derived from `toRedactedSecretValue`, so the mask has one definition and the tests cannot drift from it again.

## Verified

- `tests/unit/apis/` + `tests/unit/utils/` — **146 passed, 15 suites**, including 11 new cases covering both defects, the round-trip keep, the create-with-mask refusal, and rotation.
- `tsc --noEmit` clean for `src` and for `tests/unit/tsconfig.json`.
- `prettier --check` and `eslint` clean on the files touched.
- Six failures elsewhere in the package (`LocalSandboxProvider`, `Code Mode UDS`) are a Windows host — unix sockets and symlink paths — and **fail identically on a clean checkout with this branch stashed**, so they are not from this change.

## Notes

- `.changeset/secret-redaction-exact-mask.md` added, per the `packages/trueforge` rule in `AGENTS.md`.
- No wire-shape or schema change: the mask is a string field either way, so `packages/trueforge-sdk` and the OpenAPI spec are untouched.
- Defect 2b in the issue — the round-trip relying on the substring check — is closed by the same change, since the emitted mask is now what the keep test recognises.
- Unrelated, in case it is useful: `pnpm --filter @truefoundry/trueforge test` cannot run on a Windows host, because the script uses a `NODE_OPTIONS='…' node …` inline env prefix and cmd.exe has no such syntax. I ran jest directly to get around it. That looks like the same family as #427 and #374, so I have not touched it here.

---

> [!NOTE]
> **Medium Risk**
> Changes secret persistence and response masking for settings APIs (MCP headers, sandbox/model providers); clients may see a different mask string on GET, but round-trip keep still works for the new shapes.
> 
> **Overview**
> Tightens **settings secret redaction** so GET masks leak less and PUT “keep existing secret” only applies to the exact mask shapes the API returns—not any string that happens to contain `***REDACTED***`.
> 
> **Response masking** no longer publishes the first three characters. Long secrets (≥12 chars) now appear as `***REDACTED***-` plus the **last four** characters; shorter secrets are fully redacted.
> 
> **Write-back detection** in `isRedactedSecretValue` switches from a substring check to whole-value matching: bare sentinel or the fixed suffix-mask form from `toRedactedSecretValue`. Real secrets that embed the sentinel are stored as typed instead of silently keeping the old value. MCP header auth and sandbox provider API tests are inverted accordingly; dedicated unit tests cover masking, recognition, and `resolveStoredSecretValue`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fe3696c925d44e6b02143b3696b14f45bc5bf1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/secret-redaction-exact-mask.md
+++ b/.changeset/secret-redaction-exact-mask.md
@@ -1,0 +1,9 @@
+---
+"@truefoundry/trueforge": patch
+---
+
+Match the secret redaction sentinel exactly, and publish at most the last four characters of a stored secret.
+
+A settings write that carried `***REDACTED***` anywhere inside a real secret was treated as "keep the stored one", so the value the user typed was silently discarded — the old secret stayed, or a create failed with a message about a missing key. Only the mask a `GET` actually returns is treated as a keep now.
+
+Responses previously showed the first three and last three characters of any secret of ten characters or more, which left little of a short key unknown. A secret now masks to its last four characters and only when it is at least twelve long; anything shorter is masked completely.
--- a/packages/trueforge/src/utils/secretRedaction.ts
+++ b/packages/trueforge/src/utils/secretRedaction.ts
@@ -4,21 +4,49 @@
  */
 
 export const SECRET_REDACTION = '***REDACTED***';
-const MIN_LENGTH_FOR_PREFIX_SUFFIX = 10;
+/**
+ * Below this, a suffix would be most of the secret, so nothing is shown at all.
+ * A 10-character key used to be published as its first three and last three characters.
+ */
+const MIN_LENGTH_FOR_SUFFIX = 12;
+/** How much of a long secret a response may show, matching what other providers publish. */
+const SUFFIX_LENGTH = 4;
+const MASK_SEPARATOR = '-';
 
-/** Response mask: prefix/suffix for longer secrets; full mask when too short to hide. */
+/** Response mask: a bounded suffix for longer secrets; full mask when too short to hide. */
 export function toRedactedSecretValue(secret: string): string {
-  if (secret.length < MIN_LENGTH_FOR_PREFIX_SUFFIX) {
+  if (secret.length < MIN_LENGTH_FOR_SUFFIX) {
     return SECRET_REDACTION;
   }
-  return `${secret.slice(0, 3)}-${SECRET_REDACTION}-${secret.slice(-3)}`;
+  return `${SECRET_REDACTION}${MASK_SEPARATOR}${secret.slice(-SUFFIX_LENGTH)}`;
+}
+
+/** The prefix every suffix-bearing mask starts with, which is fixed. */
+const MASK_PREFIX = `${SECRET_REDACTION}${MASK_SEPARATOR}`;
+
+/**
+ * Whole-value test for the suffix-bearing mask.
+ *
+ * Compared by length and prefix rather than by regular expression: the sentinel is full of regex
+ * metacharacters, so a pattern built from it needs escaping to stay correct.
+ */
+function isMaskWithSuffix(value: string): boolean {
+  return value.length === MASK_PREFIX.length + SUFFIX_LENGTH && value.startsWith(MASK_PREFIX);
 }
 
 /**
  * True when the client sent a redacted stand-in from a prior GET.
+ *
+ * Matched against the exact shapes `toRedactedSecretValue` produces rather than by substring. A
+ * substring test classified any real secret that happened to contain the sentinel as "keep the
+ * stored one", so the value the user typed was dropped without an error saying so — persisted as
+ * the old secret, or failing a create with a message about a missing key.
+ *
+ * A secret whose whole value is exactly one of these shapes is still indistinguishable from a mask.
+ * Nothing in-band can separate those two; only a field saying "keep" alongside the value could.
  */
 export function isRedactedSecretValue(value: string): boolean {
-  return value.includes(SECRET_REDACTION);
+  return value === SECRET_REDACTION || isMaskWithSuffix(value);
 }
 
 /**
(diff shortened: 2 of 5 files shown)


chiragjn:
Thank you for contributing but we have decided not to go ahead with this issue. Please see https://github.com/truefoundry/trueforge/issues/423#issuecomment-5437062254

We would recommend getting maintainer approval on an issue before starting any implementation work.

PR #464, opened 2026-08-27
fix(sandbox): bound the Daytona calls that requests wait on

### LINKED ISSUE #417: Daytona snapshot registration has no request timeout — GET /capabilities, GET settings, and turn creation can hang ~5 min when Daytona stalls
### What happened?

`DaytonaSandboxProvider.registerSnapshot()` issues a raw `fetch` to the Daytona API with **no `AbortSignal` / timeout**:

```ts
// packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts (main @ 506bf5c, registerSnapshot)
response = await fetch(`${this.apiUrl}/snapshots`, {
  method: 'POST',
  headers: { Authorization: `Bearer ${this.apiKey}`, 'Content-Type': 'application/json' },
  body: JSON.stringify({ name: this.buildRef, imageName: this.imageUri }),
});
```

That call is reached from `buildImage()`, which runs — **without any timeout wrapper** — inside three request-facing paths via `checkSnapshotStatus`:

1. `GET /api/v1/capabilities` (`packages/trueforge/src/apis/capabilities.ts:35`) — this is the UI's boot/capability probe, hit on app load.
2. `GET /api/v1/settings/sandbox-providers` (`packages/trueforge/src/apis/sandboxProviders.ts:41`) — the settings page.
3. Turn creation when a fresh sandbox is needed (`packages/trueforge/src/apis/turns.ts:200`, `sandboxProvider: async ({...}) => {... checkSnapshotStatus ...}`).

The only caller that bounds the round-trip is `PUT /settings/sandbox-providers`, which wraps it in `withTimeout(provider.buildImage(), BUILD_REQUEST_TIMEOUT_MS /* 3s */)` (`sandboxProviders.ts:18,84`). The three paths above have no such cap, so a Daytona endpoint that accepts the connection but stalls the response hangs these requests for up to undici's default ~5-minute header/body timeouts — per request. During that window:

- The UI capability probe and settings page hang with no feedback.
- New turns that need a sandbox stall before execution starts.
- Each hung request keeps a server connection + event-loop task alive; a Daytona brownout turns into piled-up concurrent requests.

Related side effect worth noting while here: `checkSnapshotStatus` is invoked from **GET** handlers and can *mutate* persisted status (and re-activate an idle snapshot via `snapshot.activate`), so plain reads trigger external writes — retrying clients amplify the external call volume.

### Steps to reproduce

1. Configure a Daytona sandbox provider.
2. Simulate a stalled Daytona API for `POST {apiUrl}/snapshots` (e.g. point `apiUrl` at a TCP sink that accepts but never responds, or add an artificial delay).
3. `GET /api/v1/capabilities` or `GET /api/v1/settings/sandbox-providers` (with a snapshot in a non-ready/non-fresh state so `buildImage` is reached) — the request hangs ~5 minutes instead of failing fast.

### Expected behavior

Every outbound Daytona call should carry an explicit timeout (the codebase already has both primitives: `AbortSignal.timeout` as used by `mcpOAuthFetch`, and `withTimeout` as used by the PUT handler), and status refreshes reachable from GET/request-hot paths should reuse the same bounded budget as the PUT path.

### How are you running TrueForge?

Docker Compose

### Version

main @ 506bf5c

### Relevant logs or screenshots

No error is logged during the hang; requests simply do not complete until undici's default timeouts fire.

### Environment

OS: any
Node.js version: 22.14+

ethanstoner:
Fixes #417.

## What hangs, and where

`checkSnapshotStatus` reaches Daytona with no cap on the round-trip, from three request-handling paths:

- `GET /api/v1/capabilities` — the UI's boot probe
- `GET /api/v1/settings/sandbox-providers` — the settings page
- turn creation, when a fresh sandbox is needed (`turns.ts`)

A provider that accepts the connection and then stalls holds each of those until undici's own multi-minute defaults, once per request, each keeping a server connection and an event-loop task alive. A Daytona brownout turns into piled-up concurrent requests, and nothing is logged while it happens.

Only `PUT /settings/sandbox-providers` bounded it, at `withTimeout(provider.buildImage(), 3s)`.

## What this does

**One budget, one owner.** `BUILD_REQUEST_TIMEOUT_MS` moves out of `sandboxProviders.ts` and becomes `SANDBOX_BUILD_REQUEST_TIMEOUT_MS` next to the refresh; the PUT handler imports it. Both paths issue the same calls, and a duplicated number is one edit from disagreeing — the one-canonical-owner rule in `AGENTS.md`.

**The refresh returns the persisted status when the budget expires**, rather than throwing. All three callers are asking what the status *is*, not asking to reach Daytona:

- the capability probe already catches and fails closed, so a throw would report the sandbox unconfigured on a three-second blip
- the settings read already falls back to the record
- turn creation would take an exception it has no better answer for

A stale status serves each of them better than a hang or a 500 while the provider is unwell. A real error — rejected credentials, a failed build — still propagates, so this hides nothing except slowness. There is a test for that distinction.

**Registration carries its own request timeout.** This is the part worth a second look: `withTimeout` is a `Promise.race`, so it abandons the *result* while the request keeps running. A caller-side cap alone leaves the socket open for the full undici default, which is most of what the issue is about. `AbortSignal.timeout` on the `fetch` ends the request itself. Ten seconds there, deliberately looser than the caller's three: it is the backstop for a socket nobody is waiting on any more, not the budget a request is held to.

## Verified

The two stall cases were run against the code **with the `withTimeout` calls removed** — both hang and fail on a four-second test timeout — and pass with them. The tests fail without the fix:

```
without the fix                                   with it
× stalled status read           (4004 ms)         √ (38 ms)
× stalled reactivation          (4004 ms)         √ (4 ms)
√ real failure still surfaces                     √
√ status within the budget                        √
```

- `tests/unit/apis/` + `tests/unit/sandbox/` — **153 passed**, 4 of them new.
- `tsc --noEmit` clean for `packages/trueforge` src, its `tests/unit` project, and `packages/trueforge-core`.
- `prettier --check` and `eslint` clean on the files touched.
- Six failures elsewhere in `packages/trueforge` (`LocalSandboxProvider`, `Code Mode UDS`) are a Windows host — unix sockets and symlink paths — and **fail identically on a clean checkout**.

## Deliberately not addressed

The issue's closing note: `checkSnapshotStatus` mutates persisted status and can reactivate an idle snapshot from **GET** handlers, so plain reads drive external writes and retrying clients amplify call volume. That is real, but it is a question about where the refresh belongs rather than a timeout, and fixing it would move behaviour this change is trying to leave alone. Happy to open a separate issue or PR for it if you want it split that way.

`.changeset/daytona-request-timeouts.md` covers both published packages.

---

> [!NOTE]
> **Medium Risk**
> Changes how sandbox status refresh behaves under slow or stuck Daytona responses on several API paths; timeouts surface as errors rather than indefinite hangs, which can affect caller error handling (e.g. settings GET vs capabilities fail-closed).
> 
> **Overview**
> **`checkSnapshotStatus`** no longer waits indefinitely on Daytona when it calls **`buildImage()`** (ready snapshots that may need reactivation) or **`getImageBuildStatus()`** (non-ready builds). Both round-trips are wrapped with **`withTimeout`** from **`trueforge-core`**, using a new **`STATUS_REFRESH_TIMEOUT_MS`** of 60 seconds.
> 
> That refresh runs on request paths such as capabilities boot, sandbox settings GET, and turn creation when status must be refreshed—so a stalled Daytona provider stops holding those handlers for undici’s long default instead of hanging until the client gives up.
> 
> A patch **changeset** notes the behavior for **`@truefoundry/trueforge`** and **`@truefoundry/trueforge-core`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3474dee259398e09d6eb2ba59495a46369cb6ab7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
```diff
@@ -312,6 +320,7 @@ export class DaytonaSandboxProvider implements SandboxProvider {
           'Content-Type': 'application/json',
         },
         body: JSON.stringify({ name: this.buildRef, imageName: this.imageUri }),
+        signal: AbortSignal.timeout(SNAPSHOT_REGISTER_TIMEOUT_MS),
```
thesujai: Why is it added here? if you wrap the buildImage with `withTimeout` is should work

packages/trueforge/src/apis/sandboxProviders.ts
```diff
@@ -81,7 +79,7 @@ export function createSandboxProvidersRouter<TTransaction>(deps: SandboxProvider
           ...(locked ? { build_metadata: locked.build_metadata } : {}),
         });
         const built = toSandboxStatus(
-          await withTimeout(provider.buildImage(), BUILD_REQUEST_TIMEOUT_MS, 'sandbox buildImage'),
+          await withTimeout(provider.buildImage(), SANDBOX_BUILD_REQUEST_TIMEOUT_MS, 'sandbox buildImage'),
```
thesujai: Why is this changed from 3s to 10s ?

packages/trueforge/src/sandbox/providerUtils.ts
```diff
@@ -68,6 +75,14 @@ function sandboxStatusFromRecord(record: SandboxProviderRecord): SandboxStatus {
 // Daytona deactivates idle snapshots after 14 days; revalidate at 13 to stay a day ahead.
 const READY_REVALIDATE_INTERVAL_MS = 13 * 24 * 60 * 60 * 1000;
 
+/**
+ * Cap the Daytona round-trip so a slow or unreachable provider can't hold a request (or DB txn) open.
+ *
+ * One budget for both the settings write and the refresh below: they issue the same calls, and a
+ * refresh sits on the capability probe and on turn creation, where a stall is felt sooner.
+ */
```
thesujai: Remove comments

packages/trueforge/src/sandbox/providerUtils.ts
```diff
@@ -97,11 +112,32 @@ export async function checkSnapshotStatus({
...
+       * A refresh that cannot answer in time reports what was last persisted rather than failing
+       * the caller. This runs on the capability probe, the settings read, and turn creation, none
+       * of which are asking to talk to Daytona — they are asking for the current status, and a
+       * stale answer serves them better than a hang or a 500 while the provider is unwell.
+       */
+      logger.warn('Sandbox image status refresh timed out; using last persisted status', {
+        ...extractErrorLogFields(error),
+        tenant_id,
+      });
+      return persisted;
```
thesujai: We cannot return persisted here, sanpshot can be deactivated, yet you are still returning a `ready` status

thesujai: I would this to throw as well, because the getCapabilities caller will handle this error and it would return sandbox enabled as false

packages/trueforge/tests/unit/sandbox/providerUtils.test.ts
```diff
@@ -0,0 +1,111 @@
+// The refresh builds a real provider and then talks to Daytona. Stub the provider class itself so
```
thesujai: Remove this test file, for such a trival change tests will do us no good

packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
```diff
@@ -28,6 +28,14 @@ const BUILD_STATE_INACTIVE = 'inactive';
 const BUILD_STATE_ERROR = 'error';
 const BUILD_STATE_BUILD_FAILED = 'build_failed';
 
+/**
+ * Cap on the snapshot registration round-trip.
+ *
+ * A caller racing this against `withTimeout` abandons the promise but leaves the request running,
+ * so the socket is only released by undici's own ~5-minute default. This aborts the request itself.
+ */
+const SNAPSHOT_REGISTER_TIMEOUT_MS = 10_000;
```
thesujai: Why is this 10 s ?

thesujai (changes requested):
We wanted timeout of the calls outside the transaction block to be 1 min

packages/trueforge/src/sandbox/providerUtils.ts
ethanstoner: Right — the buildImage branch exists because a ready snapshot may have been deactivated, so returning a persisted `ready` there asserts exactly the thing the call was checking. Now throws.

ethanstoner: Done, it throws. getCapabilities catches and reports sandbox disabled; the settings GET and turn creation surface the error instead of hanging, which is the point.

packages/trueforge/src/sandbox/providerUtils.ts
ethanstoner: Trimmed.

packages/trueforge/tests/unit/sandbox/providerUtils.test.ts
ethanstoner: Removed.

packages/trueforge/src/apis/sandboxProviders.ts
ethanstoner: It was not changed — the PUT is still 3s. The rename made the diff look like more than it was, so I have reverted it: `sandboxProviders.ts` is now byte-identical to `main`, and the constant stays local to that file at 3s.

The 10s you may have seen was a separate constant in `DaytonaProvider.ts`, unrelated to this line.

packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
ethanstoner: It does free the caller, but it does not stop the request — `withTimeout` is a `Promise.race`, so the losing branch keeps running and the connection stays open until undici's own default. That is the connection pile-up the issue describes during a brownout, and it is the part a caller-side cap cannot reach.

Measured against a server that accepts and never answers, using the repo's `withTimeout` verbatim, watching the server-side socket's `close` event (Node 22.15, three runs, same result):

```
withTimeout only            caller freed  516ms | server socket closed: NEVER (still open at 3.5s)
withTimeout + AbortSignal   caller freed  515ms | server socket closed: 516ms
```

So both are doing different jobs: `withTimeout` bounds how long the request handler waits, `AbortSignal` bounds how long the socket lives.

Happy to drop it if you would still rather keep this PR to the caller-side fix — say the word and I will push that. If it stays, #417 also asked for it explicitly.

packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
ethanstoner: It was arbitrary. Now 60s, matching the longest caller budget, so the socket does not outlive anyone still waiting on it — and it no longer cuts a refresh short before that budget is up.

ethanstoner:
Pushed. Summary of what changed:

- **1 min outside the transaction.** The refresh now uses a 60s budget of its own. The settings write inside the transaction is untouched and keeps its own 3s constant — I had renamed and shared it, which made the diff read as a change to the PUT. `sandboxProviders.ts` is byte-identical to `main` again.
- **Throws on expiry** instead of returning the persisted status. You are right that returning `ready` there would assert the very thing the `buildImage` branch is checking for.
- **Comments trimmed**, **test file removed**.
- **Registration timeout is 60s**, matching the longest caller budget rather than the arbitrary 10s.

One thing I pushed back on rather than silently changing, with measurements in the thread: `withTimeout` frees the caller but does not stop the request, so the socket stays open until undici's default — that is the connection pile-up in #417 that a caller-side cap cannot reach. It is one line and I will drop it if you would rather keep this PR to the caller side.

Verified after the changes: `tests/unit/apis` 135 passed, `tsc` clean across `trueforge` src, its `tests/unit` project, and `trueforge-core`, prettier and eslint clean.

packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts
thesujai: I agree that this, but for now lets revert this change, i have done this

thesujai (approved):
Thanks @ethanstoner for your work here, really appreciate it!

Issue #416, opened 2026-08-24
TFY sandbox uploadFile fails for files larger than ~96 KiB: whole payload is base64-encoded into a single argv (MAX_ARG_STRLEN/E2BIG)


aniruddhaadak80:
### What happened?

`TFYSandboxProvider.uploadFile` encodes the entire file as base64 and embeds it in a single shell command:

```ts
// packages/trueforge-core/src/core/sandbox/provider/TFYSandboxProvider.ts (main @ 506bf5c, uploadFile)
async uploadFile(params: { sandboxId: string; remotePath: string; content: Buffer }): Promise<void> {
  const encoded = params.content.toString('base64');
  const escapedPath = shellEscape(params.remotePath);

  const result = await this.exec({
    sandboxId: params.sandboxId,
    command: `echo ${shellEscape(encoded)} | base64 -d > ${escapedPath}`,
  });
  ensureExecSuccess(result);
}
```

The full base64 blob becomes part of one `exec` command string that the sandbox server ultimately hands to a shell. On Linux, a single `execve` argument is capped by `MAX_ARG_STRLEN` (~128 KiB / 32 pages), so once `echo <base64>` grows past that limit the spawn fails with `E2BIG` / "Argument list too long" — meaning **uploads larger than roughly 96–98 KiB of binary payload fail** (base64 inflates size by 4/3), well within realistic agent usage ("analyze this CSV").

Two additional costs even below the limit:

1. Memory amplification: the file exists simultaneously as `Buffer` → base64 JS string → shell-escaped copy → decoded temp on disk, i.e. ~4–5× file size in memory plus two extra copies over the wire (JSON POST body).
2. The failure arrives as an opaque non-zero exit / infra error rather than a domain error like `SandboxFileTooLargeError`, so users get "Failed to download/upload" style messages with no hint about the size ceiling.

The local provider already solves this correctly and documents why:

```ts
/** Payload on stdin so large uploads stay off argv. */
async uploadFile(...) { // LocalSandboxProvider
  const result = await this.runSandboxCommand({
    sandboxRootPath,
    command: localSandboxUploadCommand(remotePath),
    stdin: params.content,
  });
```

### Steps to reproduce

1. Run TrueForge against a TFY sandbox provider (`type: 'tfy'`).
2. Upload any file larger than ~100 KiB through the agent flow (e.g. ask the agent to analyze a CSV slightly over 100 KB).
[...]

PR #467, opened 2026-08-27
feat(sandbox): add a self-hostable Docker sandbox provider

### LINKED ISSUE #466: Sandbox provider that can be self-hosted (and pin an image toolchain)
### What problem are you trying to solve?

There is no sandbox provider that can be fully self-hosted.

`daytona` is the only one exposed in `catalog/sandbox-catalog.yaml`, and `SandboxProviderManifestSchema` is still a single variant carrying a TODO to widen it when a second provider ships. Daytona went closed source in June 2026 and the public repo is frozen, so a current instance cannot be deployed on your own infrastructure. `LocalSandboxProvider` is implemented and tested but unreachable from configuration, and it is a host-process sandbox rather than an image-based one, so it cannot pin a toolchain.

My use case is an agent that compiles and benchmarks CUDA inside the sandbox primitive — a performance number measured somewhere other than where the kernel runs is not a measurement. That needs a reproducible image with a specific CUDA toolchain, and a GPU, and a container is currently the only sandbox that can provide either.

### Proposed solution

A container-backed sandbox provider implementing the existing `SandboxProvider` interface, with an optional `gpus` field passed through to `--gpus`.

`LocalSandboxProvider` is the natural reference to port from: same interface, same path-shaped `sandboxId`, same file-transfer and traversal concerns, with `docker exec` replacing the bwrap supervisor. The shared `sandboxProviderContractSuite` already has a branch for path-id backends, so correctness is measured against your tests rather than new ones.

### Alternatives considered

**Extending `LocalSandboxProvider` instead.** Worth saying clearly, because it is easy to assume otherwise: **bubblewrap can reach a GPU.** A CUDA kernel runs under bwrap in an unprivileged user namespace at 920.1 GB/s against 920.7 GB/s on the host — the namespace is genuinely new (`/proc/self/ns/user` differs) and genuinely unprivileged (`uid_map` is `1000 1000 1`).

Probing `cuInit` through `dlopen("libcuda.so.1")`:

| bwrap configuration | `cuInit` |
| --- | --- |
| nvidia nodes `--dev-bind`, no `/sys` | `CUDA_ERROR_OPERATING_SYSTEM` (304) |
| `--ro-bind /dev` + `/sys` | `CUDA_ERROR_NO_DEVICE` (100) |
| `--dev-bind /dev` + `/sys` | **OK, devices=1** |
| `--ro-bind /dev`, nvidia nodes `--dev-bind`, + `/sys` | **OK, devices=1** |

Two requirements: `/sys` must be mounted, and the nvidia device nodes must be writable. `/dev` itself can stay read-only.

That said, **`LocalSandboxProvider` cannot be given a GPU through its configuration**, and the reason is one level down in `@anthropic-ai/sandbox-runtime` rather than in TrueForge.

Driving `hostRun`'s `createSandbox` + `runSupervisorSession` directly, with the shipped `ALLOW_READ_BY_PLATFORM.linux` policy:

| probe inside the sandbox | result |
| --- | --- |
| `/sys` present | yes |
| `ls /dev/nvidia*` | `No such file or directory` |
| `/dev` contents | `core fd full null ptmx pts random shm stderr stdin stdout tty urandom zero` |
| `/dev` mount | a fresh `tmpfs`, over the host `devtmpfs` |
| `test -w /dev/nvidiactl` | no |
| `nvidia-smi` | `couldn't communicate with the NVIDIA driver` |

So `/sys` is already satisfied, but the device nodes are not read-only — they are absent. Listing `/dev` in `allowRead` does not surface them, because SRT mounts its own minimal `/dev` and drops `/dev/*` entries from the allowlist on purpose (`dist/sandbox/linux-sandbox-utils.js`):

```js
// Skip /dev/* paths since --dev /dev already handles them
if (normalizedPath.startsWith('/dev/')) {
```

I could not find a device-passthrough option in SRT's config surface. So getting a GPU into the local provider is not a policy change you can make in this repo — it needs SRT to grow a device allowlist first. Happy to be wrong about that if there is a surface I have missed.

That is orthogonal to self-hosting anyway, which is the part I actually need.

### Additional context

Related: #387 (E2B, `help wanted`) and #414 (ComputeSDK, closed) are both asking for non-Daytona backends.

I have a working implementation — ~810 lines plus contract, GPU and hardening tests — passing `sandboxProviderContractSuite` (4/4) and leaving the default `packages/trueforge` unit suite untouched. It streams file transfer through `docker cp`/stdin rather than base64-into-argv, so it does not reproduce #416.

Raising this as an issue first per the approval note in CONTRIBUTING.md. Happy to open the PR if you want to see it, to rework it if the shape is wrong, or to leave this as analysis if a container provider is not a direction you want to take.

Environment: Linux, Node 22.14, Docker 29.7.2 with the NVIDIA runtime, RTX 4090 (sm_89), driver 580.159.04, CUDA 13.2.

rycerzes:
## Summary

Adds a Docker sandbox provider, and widens the provider manifest to a discriminated union so a second provider can exist at all.

`daytona` is the only provider exposed in the catalog, and it went closed source in June 2026, so there is currently no way to run TrueForge fully self-hosted. This is also the only provider that can pin a toolchain: the sandbox is an image, so a workload needing a specific CUDA or Python version gets a reproducible one. `gpus` passes through to `--gpus` for workloads that need a device.

Closes #466

## Changes

- `SandboxProviderManifestSchema` becomes a discriminated union on `type`; `toDaytonaSandboxProvider` becomes `toSandboxProvider` and dispatches
- New `DockerSandboxProvider` against the existing `SandboxProvider` interface — path-shaped `sandboxId`, `docker exec`, file transfer streamed through `docker cp`/stdin rather than base64-into-argv, so it does not reproduce #416
- Optional `gpus` passed through to `--gpus`; omit for no GPU
- Optional `supportsCodeMode` on the provider interface, so a session degrades to ordinary tool calls instead of failing when a provider has no bidirectional transport
- One `docker` preset in `sandbox-catalog.yaml`

## How was this tested?

`pnpm build`, `pnpm test`, `pnpm typecheck`, `pnpm lint:ci`, `pnpm format:check` all pass. Default suites unchanged at 389 and 274 tests.

Two opt-in suites added, both needing a container runtime:

- `pnpm test:docker-sandbox:contract` — 13 tests, including your `sandboxProviderContractSuite` and sibling-escape/traversal hardening
- `pnpm test:docker-sandbox:gpu` — 3 tests, skipped via `pending()` when no GPU-capable host is present, so it is safe in CI

Verified on Linux, Node 22.14, Docker 29.7.2 with the NVIDIA runtime, RTX 4090, driver 580.159.04, CUDA 13.2: `nvidia-smi` visible inside the sandbox, and an `nvcc -arch=sm_89` compile-and-run measuring 921.2 of 1008.1 GB/s.

## Checklist

- [x] No hand-edits to generated code
- [ ] Docs / `.env.example` updated if configuration or behavior changed — not done; happy to add docs if you want this direction

---

Raised as #466 first per the approval note in CONTRIBUTING.md. Happy to rework the shape, or to close this if a container provider is not a direction you want to take.

---

> [!NOTE]
> **Medium Risk**
> Introduces container execution via the host Docker/Podman socket with new path and symlink containment logic; misconfiguration or bugs could affect host isolation, though the PR adds explicit hardening and scoped cleanup.
> 
> **Overview**
> Adds a **self-hostable Docker sandbox backend** so tenants can run agent workloads in one container per session, with optional **`--gpus`** passthrough and a configurable image (catalog preset uses `python:3.12-slim`).
> 
> **Provider model & API:** `SandboxProviderManifest` is now a **`type`-discriminated union** (`daytona` | `docker`). Runtime construction goes through **`toSandboxProvider`** instead of Daytona-only helpers; settings PUT/GET **redact and merge secrets only for Daytona**, since Docker has no stored credentials. The shipped **sandbox catalog** includes a `docker` preset.
> 
> **Code Mode:** Providers may set optional **`supportsCodeMode`**; when it is `false`, **`Sandbox.configureCodeMode` logs and skips** Code Mode instead of failing the session (Docker sets this because it has no shared-filesystem transport).
> 
> **Docker implementation highlights:** Path-shaped sandbox ids map to derived container names (works across fresh provider instances per turn). **File upload/download** uses **`docker exec` with stdin/stdout**, avoiding argv size limits. **Hardening** includes in-container timeouts, `realpath`-based path containment for file ops, scoped **stale container reaping**, and background image pulls with process-wide pull state.
> 
> **Tests:** Opt-in **`test:docker-sandbox:contract`** (shared provider contract + hardening) and **`test:docker-sandbox:gpu`** (CUDA smoke, skipped without GPU).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c22c327fcffffba6248092a09b038899c87ff1fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/docker-sandbox-provider.md
+++ b/.changeset/docker-sandbox-provider.md
@@ -1,0 +1,6 @@
+---
+'@truefoundry/trueforge-core': patch
+'@truefoundry/trueforge': patch
+---
+
+Add a Docker sandbox provider with optional GPU passthrough, widen the sandbox provider manifest to a discriminated union, and let a provider declare whether it supports Code Mode so sessions degrade instead of failing.
--- a/packages/trueforge-core/src/core/sandbox/Sandbox.ts
+++ b/packages/trueforge-core/src/core/sandbox/Sandbox.ts
@@ -267,6 +267,14 @@ export class Sandbox extends LocalToolMCP {
     if (!servers.length) {
       return;
     }
+    if (this.provider.supportsCodeMode === false) {
+      // Degrade rather than fail: the agent keeps its tools and its sandbox, and
+      // simply routes calls individually instead of batching them in a script.
+      this.logger.info('Code Mode unavailable for this sandbox provider; continuing without it', {
+        provider: this.provider.type,
+      });
+      return;
+    }
     if (this.codeModeDispatcher !== undefined || this.codeModeTransport !== undefined) {
       throw new Error('Code Mode is already configured for this Sandbox');
     }
--- a/packages/trueforge-core/src/core/sandbox/provider/Provider.ts
+++ b/packages/trueforge-core/src/core/sandbox/provider/Provider.ts
@@ -68,6 +68,15 @@ export interface SandboxProvider {
   /** Stable provider kind used in fancy sandbox ids and carry-forward (plain string). */
   readonly type: string;
   /**
+   * Whether this provider can carry Code Mode's bidirectional transport.
+   *
+   * Optional for backwards compatibility: absent means yes, which is correct for
+   * every provider that predates the flag. A provider that sets this to false is
+   * skipped rather than asked and allowed to throw -- Code Mode is an
+   * optimisation, and losing it must not fail the session.
+   */
+  readonly supportsCodeMode?: boolean;
+  /**
    * Ensures the release image is being built into the provider's backing store and
    * returns its current status. Idempotent: an already-built image reports `ready`;
    * a fresh build starts in the background and reports `pending`.
--- a/packages/trueforge/catalog/sandbox-catalog.yaml
+++ b/packages/trueforge/catalog/sandbox-catalog.yaml
@@ -7,3 +7,10 @@ providers:
     auto_stop_interval_in_minutes: 5
     auto_archive_interval_in_minutes: 60
     auto_delete_interval_in_minutes: 7200
+  - type: docker
+    # The image must provide python3 and pydantic: the sandbox bootstrap runs a
+    # Python script to materialise git-backed skills, and without them
+    # initialisation fails with "python3: not found" while the session still
+    # starts, which makes the failure easy to miss.
+    image: python:3.12-slim
+    exec_timeout_ms: 600000
--- a/packages/trueforge/jest.docker-sandbox.contract.config.cjs
+++ b/packages/trueforge/jest.docker-sandbox.contract.config.cjs
@@ -1,0 +1,43 @@
+/** @type {import('jest').Config} */
+module.exports = {
+  testEnvironment: 'node',
+  transform: {
+    '^.+\\.tsx?$': [
+      '@swc/jest',
+      {
+        jsc: {
+          parser: { syntax: 'typescript', decorators: true, dynamicImport: true },
+          target: 'es2022',
+        },
+        module: { type: 'commonjs' },
+      },
+    ],
+    '^.+\\.m?js$': [
+      '@swc/jest',
+      {
+        jsc: {
+          parser: { syntax: 'ecmascript', dynamicImport: true },
+          target: 'es2022',
+        },
+        module: { type: 'commonjs' },
+      },
+    ],
+  },
+  transformIgnorePatterns: [],
+  moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1',
+    '^@truefoundry/trueforge-core/agent-session$': '<rootDir>/../trueforge-core/src/agent-session/index.ts',
+    '^@truefoundry/trueforge-core/agent-session/(.*)$': '<rootDir>/../trueforge-core/src/agent-session/$1',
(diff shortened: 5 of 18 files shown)


rycerzes:
Flagging this on the PR as well as the issue, since the description has changed underneath anyone who already read it.

I opened this saying bubblewrap cannot expose a GPU because it unshares into a user namespace and the NVIDIA driver refuses to initialise there. That is wrong. A real CUDA kernel runs inside bwrap, in an unprivileged user namespace, at **920.1 GB/s against 920.7 GB/s on the host**. The blocker in my original testing was a missing `/sys` bind, plus device nodes that need to be writable — [full detail and the probe table here](https://github.com/truefoundry/trueforge/issues/466#issuecomment-5434642663).

So the headline reason for this PR is gone. I have retitled it around the reason that survives — `daytona` being the only exposed provider and no longer self-hostable — but that is a much weaker case than the one I opened with, and I would rather you closed this than merged it on my say-so. No objection at all to a close.

If the local-provider policy change is the more interesting thread, I think it is, and I would be glad to pick it up.

PR #477, opened 2026-08-27
fix(sandbox): report a missing Daytona snapshot permission as 403, not an invalid key

### LINKED ISSUE #461: Daytona provider setup misreports missing write:snapshots permission as an invalid API key
### What happened?

TrueForge 0.1.4 reports a valid, usable Daytona API key as invalid when the key can create/delete sandboxes but does not carry Daytona's separate `write:snapshots` permission.

I hit this while proving the sandbox path for a TrueForge hackathon project. The same key produced these direct Daytona results:

```text
GET    /api/sandbox                         -> 200
POST   /api/sandbox                         -> 200 (sandbox created)
DELETE /api/sandbox/{id}?force=true         -> 200
GET    /api/snapshots/{missing-name}        -> 404 (authenticated not-found)
POST   /api/snapshots                       -> 403 { "message": "Access denied" }
```

But a schema-valid request to:

```text
PUT /api/v1/settings/sandbox-providers
```

returns:

```text
422 Daytona rejected the API key — check the credentials
```

The credential is not rejected. Its sandbox permission works; snapshot registration is forbidden.

The cause is visible in the 0.1.4 path:

1. sandbox-provider setup calls `provider.buildImage()`;
2. `DaytonaSandboxProvider.buildImage()` registers the bundled image with `POST /snapshots` when the deterministic snapshot does not exist;
3. a Daytona `403` is classified by `isDaytonaAuthError()` alongside `401` and rewritten as a bad-key message.

Daytona's bundled SDK models `write:snapshots` separately from `write:sandboxes`, so this is an insufficient-permission result, not invalid authentication.

### Reproduction

1. Create a Daytona API key that can write/delete sandboxes but cannot write snapshots.
2. Confirm that the key can create and delete a sandbox directly.
3. Configure a Daytona sandbox provider in TrueForge 0.1.4 with that key.
4. Observe the 422 bad-credential response.
5. Send the same snapshot registration request directly and observe the underlying `403 Access denied`.

No key value is needed in logs; HTTP status and the redacted error body are sufficient.

### Expected behavior

- Distinguish `401 invalid credential` from `403 authenticated but insufficient permission`.
- Name the required Daytona permission (`write:snapshots`) in the API/UI error, or document it next to the API-key field.
- Ideally, consider a direct-image fallback when `write:sandboxes` is available but `write:snapshots` is not. With the same key, `daytona.create({ image: <TrueForge bundled image> })` successfully created a sandbox, executed `python -c "print(17*19)"` with exit 0/result 323, and deleted it. A local compatibility patch using that path also completed the full TrueForge turn:

```text
truefoundry-system exec call
-> sandbox.created (Daytona)
-> tool.response { success: true, exitCode: 0, result: "323\n" }
-> turn.done / DAYTONA_OK:323
```

If snapshot registration is intentionally mandatory, the first two bullets still prevent a valid key from being blamed for a missing scope.

### Environment

- TrueForge: 0.1.4, standalone/SQLite
- `@daytona/sdk`: 0.204.1 (bundled by TrueForge)
- Daytona API response version observed: 0.207.0
- Node.js: 24.14.1
- npm: 11.11.0
- macOS

Related but not duplicate: #417 covers missing timeouts on the snapshot-registration request. This report is about permission classification and the forced snapshot-write requirement.

ethanstoner:
Fixes #461.

## What was wrong

Daytona scopes snapshot writes separately from sandbox writes. A key that can create sandboxes but has not been granted Snapshots (write) is still a valid key, but TrueForge builds a release snapshot in the Daytona account and clones every sandbox from it, so saving the provider fails at that step. The settings write answered `422 "Daytona rejected the API key — check the credentials"`, which points the user at the one thing that is not wrong — the documented fix is to add the missing grant, not to replace the key.

## Root cause

`isDaytonaAuthError` treated 401 and 403 as a single condition:

```ts
return error instanceof DaytonaError && (error.statusCode === 401 || error.statusCode === 403);
```

The two are different answers. 401 means the credentials were not accepted, and retrying the same key cannot succeed. 403 means Daytona authenticated the key and declined the operation. Collapsing them left the PUT handler unable to distinguish "wrong key" from "key missing a permission", so both took the credentials branch.

## What changed

- `providerUtils.ts` — `isDaytonaAuthError` now matches 401 only; new `isDaytonaPermissionError` matches 403.
- `sandboxProviders.ts` — the permission case returns 403 and names `write:snapshots` in the message, so the response says which grant is missing instead of implying the key is bad.
- `sandboxProviderRoutes.ts` — `putSandboxProviderRoute` declares the 403 so the generated OpenAPI document describes it.
- Changeset added, per AGENTS.md.

## Behaviour change

The forbidden case now answers **403 where it previously answered 422**. This is the correction asked for on the issue ("The error should have been a 403 instead of a 422").

It required updating one existing assertion that expected 422 for a `DaytonaError(403)`. That test's actual subject is that config is not persisted when the build call fails; it still asserts the follow-up `GET` returns 404, so the coverage is unchanged and only the status it waits on moved. A new test covers the 403 body and message.

## Not addressed

The report's third bullet suggests falling back to running the image directly when `write:snapshots` is absent. That decides whether snapshot registration stays mandatory, which is a product question about the provider's setup contract rather than an error-mapping bug, and it would change behaviour for keys that work today. Happy to take it in a follow-up if you want it.

## Verification

- `tests/unit/apis/sandboxProviders.test.ts`: 13 passed.
- Full `@truefoundry/trueforge` unit suite: 270 passed, 6 failed. The six are `LocalSandboxProvider` and `Code Mode UDS` on a Windows host (unix sockets, and a `0700` chmod that reports `0666`). They fail identically on a clean checkout of `main`, which runs 269 passed with the same 6 failed — this branch adds exactly one passing test and changes nothing else.
- `pnpm typecheck` and `pnpm lint:ci`: clean.
- The five changed files are Prettier clean. `pnpm format:check` does fail locally, but on ~779 files including `README.md`, `SECURITY.md` and `tsconfig.base.json` that this branch never touches; it is a CRLF artifact of a Windows checkout, and the committed blobs are LF. I deliberately did not run `prettier --write` across the repo.

https://claude.ai/code/session_01XCubdWKdbhQ2vVuDdYu68E

---

> [!NOTE]
> **Low Risk**
> Limited to sandbox provider settings error mapping and messaging; no auth, persistence, or Daytona integration logic beyond how errors are classified.
> 
> **Overview**
> When saving the Daytona sandbox provider, a **valid key that lacks snapshot-related grants** used to surface the same **422** response as a bad API key (`Daytona rejected the API key`). The PR **splits Daytona error handling**: `isDaytonaAuthError` now matches **401 only**, and a new `isDaytonaPermissionError` matches **403**.
> 
> On PUT, **403 from Daytona** now returns a **422** with a message that names the grants to add in the Daytona dashboard (`write:sandboxes`, `write:snapshots`, `delete:snapshots`) instead of implying the credentials are wrong. **401** still maps to the invalid-key message. A unit test asserts the new body; the “does not persist on failure” case still uses a mocked **403** and expects **422** plus no stored config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49da9eb712bd65b3ba474a4cfd681fda7bd2c2c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/sandbox/providerUtils.ts
```diff
@@ -16,9 +16,18 @@ import {
...
+  return error instanceof DaytonaError && error.statusCode === 401;
+}
+
+/**
+ * Daytona authenticated the key but refused the operation (403 forbidden). Snapshot registration is
+ * gated on `write:snapshots`, which Daytona scopes separately from `write:sandboxes`, so a key that
+ * creates sandboxes can still fail here — the key is valid and needs a permission, not replacing.
+ */
+export function isDaytonaPermissionError(error: unknown): boolean {
+  return error instanceof DaytonaError && error.statusCode === 403;
```
thesujai, line 25: Can you see what error message Daytona returns with 401 and 403? Maybe based on that we can show the error message without doing any changes?

Some thing like `Invalid Daytona API key: {error.msg}`

This is just a thought

ethanstoner: Checked. The messages are:

- **403** — `"Access denied"`
- **401** — `"Invalid credentials"`

And the message does reach us intact, so that part of your idea works. The 403 in this issue never goes through the SDK's axios interceptor at all — `registerSnapshot` uses raw `fetch` and builds the error itself, taking the body's `message` verbatim:

```ts
// packages/trueforge-core/src/core/sandbox/provider/DaytonaProvider.ts:322-326
const message =
  isRecord(body) && typeof body['message'] === 'string'
    ? body['message']
    : `Daytona snapshot registration failed (${String(response.status)})`;
throw new DaytonaError(message, response.status);
```

The catch is what that string can ever contain. On Daytona's side the 403 is a zero-argument exception:

```ts
// daytonaio/daytona, apps/api/src/common/exceptions/access-denied.exception.ts
export class AccessDeniedException extends ForbiddenException {
  constructor() {
    super('Access denied')   // no parameter
  }
}
```

thrown by the org permission guard for `@RequiredOrganizationResourcePermissions([OrganizationResourcePermission.WRITE_SNAPSHOTS])` on `POST /snapshots` (`snapshot.controller.ts:93`). So Daytona will never name the missing permission — `"Access denied"` is a constant. The only place `write:snapshots` is knowable is our side, statically, because we know we are POSTing to `/snapshots`.

That means `Invalid Daytona API key: {error.msg}` renders, for the exact case in #461:

> `Invalid Daytona API key: Access denied` — with HTTP 422

which still tells the operator their key is invalid when it is not. Daytona authenticated it: the same key creates and deletes sandboxes, and the reporter's `GET /snapshots/{missing}` returned an authenticated 404. Only snapshot registration is forbidden. So that wording keeps the part of the bug the issue is actually about, and the status stays 422 rather than the 403 you suggested in the issue.

Your underlying point stands though — showing Daytona's own text is better than only showing ours. Both, if you want it:

```
401 → 422  Daytona rejected the API key — check the credentials (Invalid credentials)
403 → 403  Daytona accepted the API key but denied the request — the key needs the
           write:snapshots permission to register the sandbox image (Access denied)
```

That is this PR plus one interpolation per branch. Say the word and I'll push it.

One thing to flag if you were thinking of the SDK's typed subclasses: `DaytonaForbiddenError` would not work here. `STATUS_CODE_TO_ERROR` is only applied on the axios path, and the 403 above is a bare `DaytonaError` from the `fetch` path, so `instanceof` would miss it. The `statusCode === 403` check is the one that matches.

Caveat on sourcing: the 403 string is confirmed both in Daytona's source and by the reporter's live run. The 401 string is from source only — `daytonaio/daytona`'s default branch is now just a README, so the newest tag carrying source is `v0.190.0` while the bundled SDK is `0.204.1`. I have not observed a live 401.

thesujai: So better would be:
`Daytona denied access: the API key is missing required permissions. Grant write:sandboxes, write:snapshots, and delete:snapshots on the key in the Daytona dashboard, then try again.`

thesujai: Also after the changes can you incluse the SS of the error message, I want to ensure we are not breaking the UI

ethanstoner: Done, verbatim. The 403 body is now:

```
Daytona denied access: the API key is missing required permissions. Grant write:sandboxes, write:snapshots, and delete:snapshots on the key in the Daytona dashboard, then try again.
```

Changed in `packages/trueforge/src/apis/sandboxProviders.ts`, with the assertion in `tests/unit/apis/sandboxProviders.test.ts` and the changeset text updated to match. Nothing else moved — the 401/403 split, the status mapping, and the declared OpenAPI 403 are unchanged.

ethanstoner: Screenshots below, light and dark. This is the real settings UI, not a mock-up, but I should be explicit about how it was produced: I do not have a Daytona key that lacks `write:snapshots`, so I ran the standalone server (`STANDALONE=true`, SQLite) against a local stub on `127.0.0.1:8899` standing in for the Daytona API — `GET /api/snapshots/{name}` → 404, `POST /api/snapshots` → 403 `{"message":"Access denied"}`, which is what Daytona itself returns. Everything after that is real: `PUT /api/v1/settings/sandbox-providers` returns

```
HTTP/1.1 403 Forbidden
{"error":{"message":"Daytona denied access: the API key is missing required permissions. Grant write:sandboxes, write:snapshots, and delete:snapshots on the key in the Daytona dashboard, then try again."}}
```

and the browser renders it through the normal path — SDK error → `getErrorMessage` picks `body.error.message` → `SandboxSettings` `formError` → the `<p>` above Save in `ConfigureSandboxForm`.

On the UI question: the string is 181 characters and wraps to four lines inside the modal. The container is a plain `<p className="text-failure-bg text-sm">` with no width cap, clamp or truncation, so it grows the modal downward and Save stays visible without scrolling. The modal does not change width. Layout is the same in both themes.

Daytona 403 error, light theme

Daytona 403 error, dark theme

One thing worth flagging separately: two Windows-only issues turned up while getting this running, both unrelated to this PR.

1. The standalone server cannot boot on Windows. `migrateSqliteToLatest` uses kysely's `FileMigrationProvider`, which does `import(absolutePath)`; the Windows ESM loader rejects that with `Only URLs with a scheme in: file, data, and node are supported… Received protocol 'c:'`, so startup fails before the HTTP listener. It needs `pathToFileURL()`. I patched it locally only to take these screenshots and reverted the patch — happy to open a separate issue or PR for it if you want.
2. Six unit tests in `packages/trueforge` fail on a Windows host (`LocalSandboxProvider` and the Code Mode UDS transport — unix sockets and a `chmod 0700` assertion). They fail identically on a clean checkout of `main`, so they are not from this change.

packages/trueforge/src/sandbox/providerUtils.ts
```diff
@@ -16,9 +16,18 @@ import {
...
 export function isDaytonaAuthError(error: unknown): boolean {
-  return error instanceof DaytonaError && (error.statusCode === 401 || error.statusCode === 403);
+  return error instanceof DaytonaError && error.statusCode === 401;
+}
+
+/**
+ * Daytona authenticated the key but refused the operation (403 forbidden). Snapshot registration is
+ * gated on `write:snapshots`, which Daytona scopes separately from `write:sandboxes`, so a key that
+ * creates sandboxes can still fail here — the key is valid and needs a permission, not replacing.
+ */
```
thesujai:
```suggestion
```

packages/trueforge/src/sandbox/providerUtils.ts
thesujai: The windows issue is being worked on here - https://github.com/truefoundry/trueforge/pull/374

thesujai (approved):

thesujai:
@ethanstoner thanks for your work here:)

packages/trueforge/src/apis/sandboxProviders.ts
```diff
@@ -106,6 +107,17 @@ export function createSandboxProvidersRouter<TTransaction>(deps: SandboxProvider
       if (isDaytonaAuthError(error)) {
         return c.json({ error: { message: 'Daytona rejected the API key — check the credentials' } }, 422);
```
thesujai, line 108: Sorry missed this last time - can we change this to 401? Since now we are distinguishing 403 from 401

packages/trueforge/src/sandbox/providerUtils.ts
ethanstoner: Screenshots below. Both are `ConfigureSandboxForm` itself, mounted inside the real `ThemeProvider` with the package's compiled `dist/styles.css`, with `error` set to the exact string this PR returns. An API key is typed in so Save renders enabled, matching the state you land in after a failed submit. It is the component under a harness rather than a live Daytona 403 — I do not have a key with the permissions stripped to trigger a real one — so what this shows is the layout under the new string, which is what the string can affect.

**Desktop (1280x900)** — modal is 448px, message wraps to 4 lines, Save stays inside the modal.

Daytona 403 error, desktop

**Narrow (420px)** — the bottom-sheet layout, which is where a long message would break if it were going to.

Daytona 403 error, narrow viewport

Measured rather than eyeballed, on the rendered nodes:

| | desktop | narrow |
| --- | --- | --- |
| message lines | 4 | 4 |
| clipped (`scrollHeight` vs `clientHeight`) | equal, no clip | equal, no clip |
| horizontal overflow | none | none, `document.scrollWidth == 420` |
| Save button | inside modal | ends 23px above the sheet bottom |

It is a plain `<p className="text-failure-bg text-sm">` with no truncation or `nowrap`, so it wraps and pushes the button down rather than overflowing. Nothing else moves.

packages/trueforge/src/apis/sandboxProviders.ts
ethanstoner: Done in `640897b0` — 401 now, and the declared response on `putSandboxProviderRoute` moved from 422 to 401 with it, so the spec and the handler agree.

Flagging one consequence since it is wire-visible: 422 was the shipped status for this branch, so the regenerated SDK will change with it. I have not touched `docs/openapi.json` or `.github/fern/openapi/openapi.json` — `AGENTS.md` says both are CI-generated and must not be hand-edited, so I left the regen to CI.

Also updated the changeset, which still described the old 422 behaviour.

`pnpm test -- sandboxProviders` → 13/13 pass, including the renamed `PUT returns 401 when Daytona rejects the API key`. `tsc --noEmit` clean for both `src` and `tests/unit` — worth checking since Hono types the response status against the route's declared `responses`, so a mismatch there fails typecheck rather than at runtime.

thesujai: Sorry @ethanstoner it slipped my mind that we restricted 401 to only platform auth failures so UI can know auth status/standalone infra

PR #515, opened 2026-08-31
add session get-or-create-by-external-id endpoint

heerambavi1998:
## Summary
add session get-or-create-by-external-id endpoint. This allows referencing a session by a unique external ID, and will be used by schedules.
## How was this tested?
tests added.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a DB unique constraint and a new internal session mutation path with creator checks; race handling must stay correct for idempotent schedule retries.
> 
> **Overview**
> Adds an optional **tenant-unique `external_id`** on sessions (DB migration + partial unique index on Postgres/SQLite) and threads it through session records, stores, and create paths (public creates still set `external_id: null`).
> 
> **Core:** `Sessions.getByExternalId` / **`getOrCreateByExternalId`** look up or create by that key, with **`SessionExternalIdConflictError`** and a retry path when two callers race on the same id.
> 
> **API & SDK:** New authenticated internal route **`POST /internal/sessions/get-or-create-by-external-id`** (200 vs 201, same creator-only access as other session APIs), OpenAPI/Fern updates, and **`client.internal.sessions.getOrCreateByExternalId`**.
> 
> Also introduces shared **`newId()`** (lowercase ULID) for several id minting sites; schedule runtime wiring to this endpoint is not in this diff (intended follow-up per route comments).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fd9c3fe5db3d890b59e06698b5ac837b4f443e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/db/sqlite/migrations/20260831_000001_session_external_id.ts
```diff
@@ -0,0 +1,23 @@
...
+    `.execute(trx);
+  });
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.transaction().execute(async trx => {
+    await sql`DROP INDEX IF EXISTS session_external_id_uq`.execute(trx);
+    await sql`ALTER TABLE session DROP COLUMN external_id`.execute(trx);
+  });
+}
```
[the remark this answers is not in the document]
heerambavi1998: DROP COLUMN is an in-place schema edit, not a table rewrite, so it never deletes session rows and the CASCADE children are untouched.

.changeset/session-external-id.md
```diff
@@ -0,0 +1,7 @@
+---
+"@truefoundry/trueforge": patch
+"@truefoundry/trueforge-core": patch
+"@truefoundry/trueforge-sdk": patch
+---
+
+Add tenant-unique optional session `external_id` with an idempotent `POST /sessions/get-or-create-by-external-id` endpoint and SDK method.
```
debajyoti-truefoundry: I feel we should keep the internal APIs like this under a separate prefix like `/x/`. Please check what we are doing internally.

packages/trueforge/src/db/sqlite/migrations/20260831_000001_session_external_id.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Optional session `external_id`, unique within a tenant when set.
+ * NULLs are excluded so sessions without an external id do not collide.
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.transaction().execute(async trx => {
+    await sql`ALTER TABLE session ADD COLUMN external_id TEXT`.execute(trx);
+    await sql`
+      CREATE UNIQUE INDEX ${sql.raw(SESSION_EXTERNAL_ID_UQ)}
```
debajyoti-truefoundry: I do not think we should do this.

It is very easy to change the `SESSION_EXTERNAL_ID_UQ` variable name and make a mess.

In terms of the error message, we can just return a proper error message that explains what fields are included.

As of now, the user cannot give an ID, and it seems the external ID is the only field that will generate a conflict.

packages/trueforge-sdk/src/api/resources/sessions/client/Client.ts
```diff
@@ -258,6 +258,146 @@ export class SessionsClient {
...
+     *
+     * @example
+     *     await client.sessions.getOrCreateByExternalId({
+     *         agent: {
+     *             name: "name"
+     *         },
+     *         externalId: "external_id"
+     *     })
+     */
+    public getOrCreateByExternalId(
```
debajyoti-truefoundry: I do not think we should keep it in the session client. We should have an internal namespace for this.

packages/trueforge/src/db/sqlite/migrations/20260831_000001_session_external_id.ts
heerambavi1998: Done

.changeset/session-external-id.md
heerambavi1998: /x/ is deprecated. using /internal/

packages/trueforge/src/apis/sessions.ts
```diff
@@ -217,6 +218,122 @@ function checkSessionAccess({ userRef, createdBy }: { userRef: string; createdBy
...
+ * violation, then we return the winning row. An existing session is returned
+ * as-is; `agent` binds only the row this call creates.
+ */
+export async function getOrCreateSessionByExternalId(params: {
+  sessions: Sessions;
+  tenant_id: string;
+  external_id: string;
+  created_by: string;
+  agent: SessionRecord['agent'];
+}): Promise<{ session: SessionHandle; created: boolean }> {
```
debajyoti-truefoundry: Are we exposing `created` in the API response?

packages/trueforge/src/apis/sessions.ts
```diff
@@ -217,6 +218,122 @@ function checkSessionAccess({ userRef, createdBy }: { userRef: string; createdBy
...
+
+  const existing = await sessions.getByExternalId({ tenant_id, external_id });
+  if (existing !== undefined) {
+    return { session: existing, created: false };
+  }
+
+  try {
+    const session = await sessions.create({
+      tenant_id,
+      session_id: ulid().toLowerCase(),
```
debajyoti-truefoundry: Can you please create a simple util function and unify this everywhere?

packages/trueforge/src/apis/sessions.ts
```diff
@@ -217,6 +218,122 @@ function checkSessionAccess({ userRef, createdBy }: { userRef: string; createdBy
...
+ * Get-or-create the session bound to a tenant-scoped `external_id`.
+ * Idempotent and race-safe: the partial unique index guarantees at most one
+ * session per external id, so concurrent callers converge on the same row with
+ * no application-level lease.
+ *
+ * Read first; create only on a miss. A concurrent create loses with a unique
+ * violation, then we return the winning row. An existing session is returned
+ * as-is; `agent` binds only the row this call creates.
+ */
+export async function getOrCreateSessionByExternalId(params: {
```
debajyoti-truefoundry: Can I move this in the agent session itself.

packages/trueforge-core/src/agent-session/Sessions.ts
```diff
@@ -54,4 +59,21 @@ export class Sessions<
       session: record,
     });
   }
+
+  /**
+   * Returns the session bound to this tenant-scoped external id, or undefined.
+   * Read-only: does not bump last_activity_timestamp_ms.
+   */
+  async getByExternalId(
```
debajyoti-truefoundry, line 69: Let us also have get or create right here.

packages/trueforge/src/apis/sessions.ts
heerambavi1998: No just the session

debajyoti-truefoundry (approved):

PR #517, opened 2026-08-31
Dispatch schedule runs through the session and turn HTTP APIs.

heerambavi1998:
## Summary

Get-or-create a session keyed by run id, then create a turn only when that session has none, using the same SDK client in standalone and the dedicated controller.
Also removes distributed entry point as that requires auth fixes now that we call actual APIs. 

## Changes

-

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how scheduled work is executed (real turns + idempotency races with schedule API) and relies on loopback HTTP with auth disabled; distributed controller deployment is no longer shipped as a separate binary entry.
> 
> **Overview**
> **Scheduled cron runs now execute through the same session/turn HTTP path as interactive agents**, instead of logging a no-op when a run is due.
> 
> The schedule-dispatch loop uses `@truefoundry/trueforge-sdk` with `auth: false` against a configurable `baseUrl`. For each due run it **get-or-creates a session** via `internal.sessions.getOrCreateByExternalId` keyed by **run id**, then **creates a user-message turn** with the schedule task only when `listTurns` shows no existing turns—making redispatch safe after crashes or retries.
> 
> **Controller wiring is refactored**: `createController` / `runController` accept injected store, transactions, logger, and `baseUrl`; standalone `main` builds the controller with `http://localhost:${PORT}` and starts it **after** the HTTP server is listening so loopback calls succeed. The old **standalone `dist/controller` entry** and top-level Postgres bootstrap in `src/controller.ts` are removed (distributed controller process deferred until API auth is sorted).
> 
> Minor follow-ons: shared `SESSION_EXTERNAL_ID_UQ` constant for migrations/store, `sdk:build` on predev/pretest/pretypecheck, and unit tests for the dispatch client behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62f3e86ed751384f1c618e11353e1a40e0c1cca0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/controller.ts
```diff
@@ -1,70 +1,47 @@
...
+  const { scheduleStore, withTransaction, logger, baseUrl } = params;
+  return new Controller({
     loops: [
-      createScheduleDispatchLoop({
-        scheduleStore: new PostgresScheduleStore(db),
-        withTransaction: callback => db.transaction().execute(callback),
-        onTriggered: logTriggeredRun(logger),
+      scheduleDispatchLoop({
+        scheduleStore,
+        client: new TrueForge({ baseUrl, auth: false }),
```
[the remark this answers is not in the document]
heerambavi1998: This is known issue, auth will be added as follow up

packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -31,6 +32,34 @@ const SCHEDULE_DISPATCH_INTERVAL_MS = 60_000;
...
 const SCHEDULE_DISPATCH_LOOP_NAME = 'schedule-dispatch';
 
+type ScheduleRunApiClient = Pick<TrueForge, 'sessions'>;
+
+/**
+ * Hands a due schedule run to the API:
+ * 1. Get or create a session keyed by `run.id`.
+ * 2. Create a non-streaming turn only when that session has no turns.
+ *
+ * This call is hence idempotent.
```
debajyoti-truefoundry, line 42: Is this correct if you are running `executeScheduledRun` concurrently?

packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -31,6 +32,34 @@ const SCHEDULE_DISPATCH_INTERVAL_MS = 60_000;
...
+ */
+function executeScheduledRun(client: ScheduleRunApiClient): (item: ScheduleDispatchItem) => Promise<void> {
```
debajyoti-truefoundry, line 44: Is is possible to make this a simple function? This seems overly complicated.

packages/trueforge/src/controller.ts
```diff
@@ -94,7 +68,6 @@ try {
   process.on('SIGINT', signal => {
     void shutdown(signal);
   });
-} catch (error) {
-  console.error('Failed to start controller:', error instanceof Error ? error.message : error);
-  process.exit(1);
+
+  return controller;
```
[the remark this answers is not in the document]
heerambavi1998: THis is intentional for distributed mode, as added in PR description

packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -31,6 +32,34 @@ const SCHEDULE_DISPATCH_INTERVAL_MS = 60_000;
...
+  return async ({ run, schedule }) => {
```
debajyoti-truefoundry, line 45: Given that this function is returning void, how would I later filter sessions for a particular run?
Are we not supporting it? Can we add this feature in future? Do you see a path?

packages/trueforge/src/controller/scheduleDispatch.ts
heerambavi1998: The signature helps with the tests - where we can pass another onTriggered method - to prevent having to call session / turn creation in contracts.
Any recommendations on a simpler approach?

packages/trueforge/src/controller/scheduleDispatch.ts
heerambavi1998: No, the controller loop level assumption is no concurrent runs. Can add the same comment here as well if needed

packages/trueforge/src/controller/scheduleDispatch.ts
heerambavi1998: I did not understand this requirement. Session filtering will be separate? Why does dispatch need filtering?

packages/trueforge/package.json
```diff
@@ -62,6 +62,7 @@
     "@hono/zod-openapi": "^1.6.1",
     "@modelcontextprotocol/sdk": "^1.30.0",
     "@truefoundry/trueforge-core": "workspace:*",
+    "@truefoundry/trueforge-sdk": "workspace:*",
```
debajyoti-truefoundry, line 68: Can you check with @chiragjn on how will this affect release.

debajyoti-truefoundry (approved):

packages/trueforge/src/controller/scheduleDispatch.ts
debajyoti-truefoundry: Discussed offline.

packages/trueforge/src/controller/scheduleDispatch.ts
debajyoti-truefoundry: We will modify/remove this comment as it depends on the caller.

packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -31,6 +32,34 @@ const SCHEDULE_DISPATCH_INTERVAL_MS = 60_000;
...
+    const turns = await client.sessions.listTurns(session.id, { limit: 1 });
+    if (turns.data.length > 0) {
+      return;
+    }
+
+    await client.sessions.createTurn(session.id, {
+      input: [{ type: 'user.message', content: schedule.manifest.task }],
+      previousTurnId: 'none',
+    });
+  };
```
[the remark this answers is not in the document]
heerambavi1998: this is ok for now, will be fixed when we add retries

PR #538, opened 2026-09-01
feat: API to trigger a manual run

sr07asthana:
## Summary

## Changes

-

## Checklist

---

> [!NOTE]
> **Medium Risk**
> New endpoint starts real agent turns in-process and shares turn-execution paths with sessions; failures and RBAC are handled but behavior affects scheduling and agent execution.
> 
> **Overview**
> Adds **POST `/api/v1/schedules/runs`** so callers can start a schedule immediately with `{ schedule_id }`, using the task from the schedule manifest. The run is recorded with a `manual-*` name and `triggered` status; the existing cron **pending** run is left unchanged.
> 
> The handler enforces the same creator-scoped access as other schedule APIs, persists the run, then kicks off work in-process via shared **`startScheduleRun`** (session get-or-create by run id, idempotent turn when the session has no turns). Turn startup is factored into **`beginTurnExecution`**, **`startTurnInProcess`**, and **`getTurnExecutionError`** so the HTTP turns API and run-now path share one engine; failed starts mark the run `failed` and map errors to 400/404/422 as appropriate.
> 
> OpenAPI, Fern-generated SDK **`schedules.createRun`**, and unit/wire tests cover the new surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebb53c625e8339a5b035a303eb9fe1cddcb848ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/fern/openapi/openapi.json
```diff
@@ -980,6 +980,23 @@
         ],
         "type": "object"
       },
+      "CreateManualScheduleRunRequest": {
```
heerambavi1998: lets not have both body and path. Since its post call lets just have body then

packages/trueforge/src/apis/schedules.ts
```diff
@@ -144,6 +156,85 @@ export function createSchedulesRouter<TTransaction>(deps: SchedulesRouterDeps<TT
...
+        switch (error.code) {
+          case 'invalid_file_input':
+            return c.json({ error: { message: error.message } }, 400);
+          case 'invalid_send_input':
+          case 'agent_sandbox_required':
+          case 'tool_name_collision':
+            return c.json({ error: { message: error.message } }, 422);
+          case 'capability_state_error':
+          case 'mcp_connection_failed':
+            throw error;
```
heerambavi1998: do we need to handle this here? I am assuming same status code are returned elsewhere based on error type? Can we have a helper?

heerambavi1998: mcp connection is expected to be done in test runs - 

Test run is exactly to see if this stuff works so instead of throwing errors I believe we should return a success or error response as part of the API response with a 20x for these errors ?

packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -60,6 +73,56 @@ function executeScheduledRun(client: ScheduleRunApiClient): (item: ScheduleDispa
   };
 }
 
+/**
```
heerambavi1998, line 76: Is there a way to unify this with executeScheduledRun - by having a separate implementation  of ScheduleRunApiClient? That makes it easy to keep internal logic same

packages/trueforge/src/apis/turns.ts
```diff
@@ -334,6 +350,115 @@ export async function drainTurnEvents(input: {
...
+  turnId: string;
+  maxExecutionTimer: NodeJS.Timeout;
+  logger: Logger;
+}
+
+/**
+ * Shared create-turn engine: persist the turn, start execution, and return the
+ * drain inputs. Does not wait for events and does not write HTTP/SSE.
+ */
+export async function beginTurnExecution(params: {
```
heerambavi1998, line 367: should we move some of this logic to turn store ? Can you see what is the right boundary here?

sr07asthana: Will discuss and do in a separate PR

heerambavi1998 (approved):

PR #540, opened 2026-09-01
Add agent metadata column and store update API

bhaveshpatel640:
## Summary

Adds the AGE-2063 foundation for agent metadata: persist an `agent.metadata` jsonb column on agent registry rows and expose store-level `updateAgentMetadata`.

Linear: AGE-2063

## Changes

- PG + SQLite migrations: `agent.metadata` jsonb / JSONB blob, default `{}`
- `AgentMetadataSchema` (strict empty whitelist for now) + `IAgentStore.updateAgentMetadata`
- PG/SQLite stores read/write `metadata` on agent records
- Patch changeset for `@truefoundry/trueforge`

## How was this tested?

- Store contract: SQLite `agent-store` suite (metadata on create/update round-trip)
- `pnpm exec tsc --noEmit` (package) as needed while iterating

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Schema migrations touch the `agent` table (SQLite rebuild with FK handling); behavior change to `updateAgent` could affect internal callers, though the public API surface is unchanged.
> 
> **Overview**
> Adds an internal **`agent.metadata`** JSON column on registry rows in **Postgres** and **SQLite** (migrations default existing rows to `{}`), with a typed **`AgentMetadata`** placeholder that is empty until keys are whitelisted.
> 
> **`IAgentStore.updateAgent`** is now a **partial patch**: callers may supply **`manifest`**, **`metadata`**, or both; omitted fields stay unchanged, and create still seeds metadata as `{}`. Postgres/SQLite stores read and write the new column accordingly.
> 
> The **public Agent HTTP API is unchanged**—responses still omit metadata, and strict request schemas **reject `metadata` on PUT**. Contract and router tests cover metadata round-trips at the store layer and wire isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00c45de0a1fd3ae7ffaa8d7620ee93be1cb5e2d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/db/postgres/migrations/20260901_000001_agent_metadata.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Agent registry metadata jsonb column.
+ * Default `{}` for existing rows.
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await sql`SET LOCAL lock_timeout = '5s'`.execute(db);
+  await sql`
+    ALTER TABLE agent
+      ADD COLUMN metadata jsonb NOT NULL DEFAULT '{}'::jsonb
```
heerambavi1998, line 11: do we need to keep { } default ? why not null?

packages/trueforge/src/db/sqlite/agent-store/SqliteAgentStore.ts
```diff
@@ -79,6 +83,7 @@ export class SqliteAgentStore implements IAgentStore<Transaction<Database>> {
           tenant_id: input.tenant_id,
           name: input.name,
           manifest: jsonbBind(input.manifest),
+          metadata: jsonbBind(EMPTY_AGENT_METADATA),
```
heerambavi1998, line 85: same why explicitly set as empty? any reason?

packages/trueforge/src/db/agentStore.ts
```diff
@@ -67,6 +75,7 @@ export interface IAgentStore<TTransaction = never> {
   createAgent(input: CreateAgentInput, transaction?: TTransaction): Promise<AgentRecord>;
   /** Replaces `manifest` for an existing id. Returns undefined if missing. */
   updateAgent(input: UpdateAgentInput, transaction?: TTransaction): Promise<AgentRecord | undefined>;
+  updateAgentMetadata(input: UpdateAgentMetadataInput, transaction?: TTransaction): Promise<AgentRecord | undefined>;
```
heerambavi1998: why not just have updateAgent also take metadata. Don't need separate method here?

packages/trueforge/src/schemas/agentMetadata.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Agent-row metadata jsonb with a strict whitelist (empty until keys are added).
+ */
+import { z } from '@hono/zod-openapi';
+
+/** `agent.metadata` jsonb; unknown keys are rejected until whitelisted. */
+export const AgentMetadataSchema = z.object({}).strict().openapi('AgentMetadata');
```
heerambavi1998: we should  not need zodiac schema since its internal only

packages/trueforge/tests/db/agentStoreContractSuite.ts
```diff
@@ -78,6 +80,17 @@ export function runAgentStoreContractSuite(getStore: () => IAgentStore): void {
     expect(await store.getAgent({ tenant_id: TENANT, name: 'research' })).toEqual(updated);
   });
 
+  it('updateAgentMetadata returns undefined for unknown ids', async () => {
```
heerambavi1998: can we add happy path tests

packages/trueforge/src/db/postgres/migrations/20260901_000001_agent_metadata.ts
bhaveshpatel640: Kept it for consistency, this ensure we don't need to do checks like `metadata ?? {}`.

packages/trueforge/src/db/sqlite/agent-store/SqliteAgentStore.ts
bhaveshpatel640: Kept it for consistency, this ensure we don't need to do checks like metadata ?? {}.

packages/trueforge/src/db/agentStore.ts
bhaveshpatel640: Kept separate so public `updateAgent` stays manifest-only. Since metadata is internal, a dedicated method is explicit without partial update manifest & metadata support in updateAgent

heerambavi1998: but this method is not public. APIs are public. These are internal methods. Lets keep single method only on store.

heerambavi1998 (approved):

packages/trueforge/src/db/agentStore.ts
bhaveshpatel640: Handled here, https://github.com/truefoundry/trueforge/pull/540/commits/f2d753cf15129b2b748e10f00a8a1f37f55eeed4

PR #554, opened 2026-09-02
feat: Add POSTGRES_SSL_MODE support for Postgres connection

thesujai:
## Summary

feat: Add POSTGRES_SSL_MODE support for Postgres connection

## Changes

feat: Add POSTGRES_SSL_MODE support for Postgres connection
## How was this tested?
NA

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how external/managed Postgres connections are established when SSL is enabled; misconfigured `sslmode` can prevent startup or DB connectivity, though the default remains unchanged.
> 
> **Overview**
> Adds optional **`POSTGRES_SSL_MODE`** so distributed deployments can set Postgres TLS via the standard `sslmode` query parameter when the app builds the URL from `POSTGRES_*` (unset leaves behavior unchanged—no `sslmode` on the connection string). **`DATABASE_URL` is still used as-is** when set; this knob applies to the composed `postgres://…` path.
> 
> The server validates allowed modes (`disable`, `prefer`, `require`, `verify-ca`, `verify-full`, `no-verify`) at boot and fails fast on invalid values. Helm gains **`externalPostgres.sslMode`**, which sets `POSTGRES_SSL_MODE` for external Postgres only; docs are updated in **`.env.example`**, **`values.yaml`**, and a patch changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58aaff3757b5ec20cbcb70043eb9d27b059bf180. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/config.ts
```diff
@@ -18,6 +18,8 @@ import { fileURLToPath } from 'node:url';
 
 import envPaths from 'env-paths';
 
+import { resolvePostgresSslConfig, type PostgresSslConfig } from './db/postgres/ssl';
```
chiragjn: please keep this function in config.ts itself, we do not want config.ts importing other modules

chiragjn: <strike>

or rather don't do any resolution in config.ts
Just keep it simple POSTGRES_SSL_MODE
and let createDB resolve it

I would prefer this ^

</strike>

chiragjn: on another read, config.POSTGRES_SSL should not be a variable of its own, rather resolvePostgresDatabaseUrl should include all this logic and add the correct `sslmode` to connection string itself

packages/trueforge/src/db/postgres/ssl.ts
```diff
@@ -0,0 +1,19 @@
...
+    case 'verify-ca':
+    case 'verify-full': {
+      return true;
+    }
+    case 'no-verify': {
+      return { rejectUnauthorized: false };
+    }
+    default: {
+      return false;
+    }
```
chiragjn: We can throw an error here

chiragjn (approved):

PR #567, opened 2026-09-03
feat: per-MCP-server request headers via x-tfg-mcp-headers

pranjaltrvd:
## Why

A caller authenticates a whole turn as one identity, and since #555 that identity is what every MCP connection presents as its gateway Bearer. That is correct for the gateway, which authorises `USE_MCP_SERVER` on the caller — but it means a server that resolves permissions for the *end user behind* the caller has no way to learn who that is. Every user's tool calls arrive as the same account.

## What

`x-tfg-mcp-headers` maps an MCP server name to the headers that server's connection should carry:

```http
x-tfg-mcp-headers: {"tfy-platform-mcp": {"x-tfy-mcp-headers": "{\"Authorization\":\"Bearer <end user token>\"}"}}
```

`TrueFoundryMcpServerStore.resolveInvokeHeaders` merges the entry for the server it is resolving. Building on #555 means this is a small change — that PR moved invoke headers onto the store and already passes the record, so the server name needed to key the lookup was there.

## Notes for review

- **The gateway Bearer is written last**, so an override cannot replace the token authorisation depends on. Covered by a test.
- **A malformed value is a 400, not a no-op.** Ignoring it would silently run the end user's tool calls under the caller's own identity, which is the thing the header exists to prevent. Absent is still fine and means no overrides.
- **No server sees another server's entry** — the merge is keyed on `record.name`, which `toRecord` sets from the ServiceFoundry server name. Also covered by a test.
- Read where the request-scoped store is built, next to the existing `requireAccessToken(c)`. Not declared in OpenAPI, matching how `Last-Event-Id` is handled today; the changeset bumps only `@truefoundry/trueforge` for the same reason.
- Not added to `mcpServerStoreContractSuite` — that covers the DB-backed stores, and this behaviour is specific to `TrueFoundryMcpServerStore`.

## Open question

`x-tfg-` is a new prefix; nothing else in the repo uses it, and there is no `x-trueforge-` either. Happy to rename before this sets a precedent for the integration contract.

## Test plan

- [x] `tests/unit/truefoundry/perServerMcpHeaders.test.ts` — parsing, rejection cases, per-server isolation, and the non-overridable Bearer
- [x] Adjacent suites still pass: `getMcpConnection`, `mapSfyMcpServers`
- [x] `typecheck`, `eslint`, `prettier --check` clean
- [ ] Not verified: whether these headers can reach persisted turn events or logs. They carry a bearer token, so worth a second pair of eyes.

### DIFF
--- a/packages/trueforge/src/main.ts
+++ b/packages/trueforge/src/main.ts
@@ -83,6 +83,7 @@ import { PACKAGE_VERSION } from './packageVersion';
 import { ActiveTurnRegistry } from './runtime/activeTurns';
 import { EventSubscriptionRegistry } from './runtime/event-subscription';
 import { printStandaloneStartupBanner } from './startupBanner';
+import { parsePerServerMcpHeaders, X_TFG_MCP_HEADERS } from './truefoundry/perServerMcpHeaders';
 import { TrueFoundryMcpServerStore } from './truefoundry/TrueFoundryMcpServerStore';
 import { TrueFoundryModelProviderStore } from './truefoundry/TrueFoundryModelProviderStore';
 import { TrueFoundryServiceFoundryServerClient } from './truefoundry/TrueFoundryServiceFoundryServerClient';
@@ -163,7 +164,11 @@ function buildResolveMcpServerStore<TTransaction>(options: {
   });
   return c =>
     c
-      ? new TrueFoundryMcpServerStore<TTransaction>({ client, accessToken: requireRequestCredentialToken(c) })
+      ? new TrueFoundryMcpServerStore<TTransaction>({
+          client,
+          accessToken: requireRequestCredentialToken(c),
+          perServerHeaders: parsePerServerMcpHeaders(c.req.header(X_TFG_MCP_HEADERS)),
+        })
       : withAuthPersistence;
 }
 
--- a/packages/trueforge/src/main.ts
+++ b/packages/trueforge/src/main.ts
@@ -162,14 +162,17 @@ function buildResolveMcpServerStore<TTransaction>(options: {
     logger: options.logger,
     tls: { enabled: configuration.TRUEFOUNDRY_MTLS_ENABLED, dir: configuration.TRUEFOUNDRY_MTLS_CERTS_DIR },
   });
-  return c =>
-    c
-      ? new TrueFoundryMcpServerStore<TTransaction>({
-          client,
-          accessToken: requireRequestCredentialToken(c),
-          perServerHeaders: parsePerServerMcpHeaders(c.req.header(X_TFG_MCP_HEADERS)),
-        })
-      : withAuthPersistence;
+  return c => {
+    if (!c) {
+      return withAuthPersistence;
+    }
+    const rawPerServerHeaders = c.req.header(X_TFG_MCP_HEADERS);
+    return new TrueFoundryMcpServerStore<TTransaction>({
+      client,
+      accessToken: requireRequestCredentialToken(c),
+      perServerHeaders: rawPerServerHeaders ? parsePerServerMcpHeaders(rawPerServerHeaders) : {},
+    });
+  };
 }
 
 /** SQLite stores; Redis unused (executor peering disabled). */
--- a/packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
+++ b/packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
@@ -27,8 +27,8 @@ export function parsePerServerMcpHeaders(raw: string | undefined): PerServerMcpH
   let decoded: unknown;
   try {
     decoded = JSON.parse(raw);
-  } catch {
-    throw new HTTPException(400, { message: `${X_TFG_MCP_HEADERS} must be a JSON object` });
+  } catch (error) {
+    throw new HTTPException(400, { message: `${X_TFG_MCP_HEADERS} must be a JSON object`, cause: error });
   }
 
   const parsed = PerServerMcpHeadersSchema.safeParse(decoded);
--- a/packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
+++ b/packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
@@ -1,0 +1,41 @@
+import { HTTPException } from 'hono/http-exception';
+import { z } from 'zod';
+
+/**
+ * Per-MCP-server request headers, keyed by server name.
+ *
+ * TrueForge authenticates the whole turn as one caller, so every MCP connection would otherwise
+ * present that same identity. This carries the identity each individual server should see —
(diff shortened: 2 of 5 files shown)


sr07asthana (approved):

sr07asthana:
LGTM just confirm on the naming once, x-tfg-* or x-trueforge-*

packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
```diff
@@ -0,0 +1,41 @@
...
+const PerServerMcpHeadersSchema = z.record(z.string().min(1), z.record(z.string().min(1), z.string()));
+
+export type PerServerMcpHeaders = z.infer<typeof PerServerMcpHeadersSchema>;
+
+/**
+ * Absent is fine and means "no overrides". Malformed is not: the alternative to failing is running
+ * a user's tool calls under the caller's own identity, which is the thing these headers exist to
+ * prevent, so a bad value is rejected rather than dropped.
+ */
+export function parsePerServerMcpHeaders(raw: string | undefined): PerServerMcpHeaders {
```
heerambavi1998: why are we allowing undefined at all?

sr07asthana: check if we have a common logic for zod parsing schema. Ideally this should be a common function?

packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (raw === undefined || raw.length === 0) {
+    return {};
+  }
+
+  let decoded: unknown;
+  try {
+    decoded = JSON.parse(raw);
+  } catch (error) {
+    throw new HTTPException(400, { message: `${X_TFG_MCP_HEADERS} must be a JSON object`, cause: error });
```
heerambavi1998, line 27: why throw HTTP errors in internal methods. Caller should catch and add HTTP codes as required. If caller is a server method, than handler should catch. I

Since  there is a single caller, remove the method altogether and add this logic in resolveInvokeHeaders itself. Tests should be updated accordingly.

packages/trueforge/src/main.ts
```diff
@@ -143,7 +144,11 @@ function buildResolveMcpServerStore<TTransaction>(options: {
   });
   return c =>
     c
-      ? new TrueFoundryMcpServerStore<TTransaction>({ client, accessToken: requireAccessToken(c) })
+      ? new TrueFoundryMcpServerStore<TTransaction>({
+          client,
+          accessToken: requireAccessToken(c),
+          perServerHeaders: parsePerServerMcpHeaders(c.req.header(X_TFG_MCP_HEADERS)),
```
heerambavi1998: so header value is expected to be key value pairs?  Why do we need to nest?

why not top level headers?
```
x-tfg-1 : val1,
x-tfg-2: val2 
```

packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
pranjaltrvd: Because the header is optional — `c.req.header(X_TFG_MCP_HEADERS)` returns `string | undefined`, so absence has to be absorbed somewhere. Handling it here keeps the single call site a plain expression rather than a `?? ''`. Happy to push it to the caller if you prefer it explicit there.

pranjaltrvd: Checked — there is no shared helper today. There are 8 direct `safeParse` call sites and they each handle failure differently: this one 400s, `TrueFoundryServiceFoundryServerClient` 424s after logging the zod issues, others just fall back to a default. A common function would have to be parameterised over status, logging and message, so it felt like a repo-wide refactor rather than something to introduce in this PR. Happy to file it as a follow-up if you want it.

packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
pranjaltrvd: Two things here, and I disagree on both — but tell me if you still want it and I'll change it.

On throwing `HTTPException` from a non-route file: that is the existing convention in this directory. `TrueFoundryServiceFoundryServerClient` throws `HTTPException(424)` directly off a failed `safeParse`, and `TrueFoundryMcpServerStore.managed()`, `TrueFoundryModelProviderStore` and `mapEnabledModels` all throw `HTTPException` too. None of them are route handlers.

On folding it into `resolveInvokeHeaders`: that method is on the shared `IMcpServerStore` interface with four implementations (Sqlite, Postgres, `McpServerWithAuthStore`, this one). It is synchronous and only ever receives a `McpServerRecord` — it never sees the raw header. To parse there I would have to pass the unparsed string into the constructor and parse on every call. It is reached from `getMcpConnection`, once per MCP connection during a turn, so a malformed header would become a mid-turn tool failure instead of a rejected request. Parsing once at the edge is what lets a bad value be a 400.

packages/trueforge/src/main.ts
pranjaltrvd: The nesting is what carries the server name, and routing a header to exactly one server is the point of the feature — the end-user token meant for `tfy-platform-mcp` must not be sent to a third-party MCP server the agent also has access to. Flat `x-tfg-1: val1` has nowhere to say which server a value belongs to, so every server would see all of them.

A flat form that kept the isolation would have to encode the server into the header name (`x-tfg-mcp-<server>-<header>`), which is ambiguous to split once a server name contains a dash and breaks for names that are not valid header-name characters.

It also matches what already exists on the way in: sfy-server and the AI Gateway pass `x-tfy-mcp-headers` as a JSON object in a single header, so a JSON object keyed by server is the consistent shape.

pranjaltrvd:
Naming: sticking with `x-tfg-*`. There is no precedent either way in this repo — the only custom headers today are `x-fern-*` (SDK codegen annotations) and `x-excluded` — so this establishes one. `x-tfg-` pairs with TrueFoundry's existing `x-tfy-` prefix, and both appear on the same request: sfy-server sends `x-tfg-mcp-headers` to TrueForge, whose values carry `x-tfy-mcp-headers` onward to the gateway. `tfy` = TrueFoundry, `tfg` = TrueForge. Shout if you would rather have it spelled out as `x-trueforge-`.

Also: both Bugbot findings were fixed in b99152ea (case-insensitive authorization drop, and `cause` preserved on the parse rethrow) — that review ran against the earlier f6b98593.

heerambavi1998 (approved):
LG

packages/trueforge/src/truefoundry/perServerMcpHeaders.ts
heerambavi1998: prefer explicit here

PR #573, opened 2026-09-03
[DO NOT MERGE] Authorizer POC

chiragjn:


packages/trueforge/src/auth/externalAuthorizer.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  constructor(input: { client: TrueFoundryAgentAuthorizationClient }) {
+    this.#client = input.client;
+  }
+
+  async listAgentAccess(input: {
+    context: AgentAuthorizationContext;
+    action: AgentAction;
+  }): Promise<ExternalAgentListAccess> {
+    if (input.context.is_admin) {
```
debajyoti-truefoundry, line 60: This will go away.

packages/trueforge/src/auth/externalAuthorizer.ts
```diff
@@ -0,0 +1,125 @@
...
+      }),
+    };
+  }
+
+  async canAccessAgent(input: {
+    context: AgentAuthorizationContext;
+    action: AgentAction;
+    agent: AgentRecord;
+  }): Promise<boolean> {
+    if (input.context.is_admin) {
```
debajyoti-truefoundry, line 77: This will go away.

packages/trueforge/src/auth/externalAuthorizer.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    return this.#client.canAccessAgent({
+      subject: credentialSubjectFor(input.context),
+      action: input.action,
+      agent_external_id: requireExternalId(input.agent),
+    });
+  }
+}
+
+export function listAccessibleAgents(input: {
```
debajyoti-truefoundry, line 88: How does pagination work here?

packages/trueforge/src/auth/externalAuthorizer.ts
```diff
@@ -0,0 +1,125 @@
...
+    tenant_id: input.tenant_id,
+    external_ids: input.access.kind === 'all' ? undefined : input.access.agent_external_ids,
+  });
+}
+
+/**
+ * Visibility for owner-or-agent list queries, with external agent ids mapped to
+ * internal ones so the session and schedule stores never see external ids.
+ */
+export async function resolveListVisibility(input: {
```
debajyoti-truefoundry, line 103: Can we please keep this in the controller / router or service functions now?

packages/trueforge/src/db/postgres/schedule-store/PostgresScheduleStore.ts
```diff
@@ -228,9 +230,17 @@ export class PostgresScheduleStore implements IScheduleStore<Transaction<Databas
     if (input.agent_names !== undefined) {
       query = query.where('agent_name', 'in', [...input.agent_names]);
     }
-    if (input.created_by_subject_id !== undefined) {
-      query = query.where(sql`created_by_subject->>'subject_id'`, '=', input.created_by_subject_id);
-    }
+    query = query.where(eb => {
+      const { owner_subject_id, agents } = input.visibility;
```
debajyoti-truefoundry, line 234: Can we just keep this flat?

agent ids will be a simple filter and `created_by_subject_id` will also be a simple filter.

packages/trueforge/src/db/postgres/schedule-store/PostgresScheduleStore.ts
```diff
@@ -228,9 +230,17 @@ export class PostgresScheduleStore implements IScheduleStore<Transaction<Databas
...
+      const owned = eb(sql<string>`created_by_subject->>'subject_id'`, '=', owner_subject_id);
+      if (agents.kind === 'all') {
```
debajyoti-truefoundry, line 236: Let us not introduce these concepts in store. Keep store a simple query interface please.

packages/trueforge/src/db/postgres/schedule-store/PostgresScheduleStore.ts
debajyoti-truefoundry: I understand the `or` condition but see if we can just keep this a simple query interface. If we cannot then we will see.

packages/trueforge/src/db/postgres/session-metrics/queries.ts
```diff
@@ -45,7 +45,9 @@ async function fetchSessionMetricsAggregate(
     ])
     .where('tenant_id', '=', input.tenant_id)
     .where('agent_id', '=', input.agent_id)
-    .where(sql`created_by_subject->>'subject_id'`, '=', input.created_by_subject_id)
+    .$if(input.created_by_subject_id !== undefined, qb =>
```
debajyoti-truefoundry, line 48: Is metrics no visible to just agent manager? If yes, do we need this?

PR #584, opened 2026-09-04
[AuthZ: 1/N] Remove TrueFoundry global admin

chiragjn:
> [!NOTE]
> **Medium Risk**
> Changes who counts as admin in TrueFoundry mode (settings/capabilities and schedule admin visibility), with no replacement tenant-wide admin in this PR.
> 
> **Overview**
> **Removes the temporary `tenant-admin` → TrueForge admin mapping** in TrueFoundry mode as the first step in a broader AuthZ rollout.
> 
> `hasAdminRole` now always returns **false** when `getTrueForgeMode()` is TrueFoundry, and the `TRUEFOUNDRY_ADMIN_ROLE` constant is deleted. Docs and a rename TODO note that this gate is still used for settings, capabilities, and schedule bypass until `canAccessSettings` lands.
> 
> **OIDC admin checks are tightened:** instead of falling back to the standalone `admin` role when OIDC config is missing, the OIDC branch returns false unless OIDC is configured and `OIDC_ADMIN_ROLE_VALUE` is present in `roles`. Standalone behavior is unchanged (`admin` role).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92e4ed3042ae277d331f825bc423c7c49e17599a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/trueforge/src/auth/identity.ts
+++ b/packages/trueforge/src/auth/identity.ts
@@ -3,9 +3,6 @@ import type { Context } from 'hono';
 
 import configuration, { getTrueForgeMode, isOidcConfigured, TrueForgeMode } from '../config';
 
-/** TrueFoundry Control Plane role treated as admin until mapping is settled. */
-export const TRUEFOUNDRY_ADMIN_ROLE = 'tenant-admin';
-
 /** Standalone / default TrueForge admin role string. */
 export const STANDALONE_ADMIN_ROLE = 'admin';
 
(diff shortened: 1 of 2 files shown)


packages/trueforge/src/auth/identity.ts
```diff
@@ -55,13 +52,12 @@ export function resolveRequestContext(c: Context): RequestContext {
  * Mode and OIDC admin claim value come from process config ({@link getTrueForgeMode}).
  * - Standalone: `roles` includes `admin`
  * - OIDC: `roles` includes configured `OIDC_ADMIN_ROLE_VALUE`
- * - TrueFoundry: `roles` includes `tenant-admin` (temporary; revisit mapping later)
+ * - TrueFoundry: no tenant-wide TrueForge admin
  */
 export function hasAdminRole(requestContext: Pick<RequestContext, 'roles'>): boolean {
```
debajyoti-truefoundry, line 58: If the ONLY thing this is guarding is the settings page, i would like to reduce all confusion and call it:


```suggestion
export function canAccessSettingsPage(requestContext: Pick<RequestContext, 'roles'>): boolean {
```

debajyoti-truefoundry (approved):

packages/trueforge/src/auth/identity.ts
chiragjn: Renaming will be done at the end, currently schedule rbac uses this function
once that is replaced by authorizer this will not be needed

PR #591, opened 2026-09-04
feat: add session metadata filter

sr07asthana:
## Summary

## Changes

-

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new list API query contract, custom deepObject parsing, and a Postgres GIN migration; behavior is well-tested but changes how clients must encode metadata filters.
> 
> **Overview**
> **List sessions** can now be filtered by exact metadata containment using OpenAPI **deepObject** query params (`metadata[key]=value`). Sessions must include all supplied key/value pairs; extra metadata on the session is allowed. An empty folded map (`metadata={}` via no bracket params) is treated as no filter at the HTTP layer, while the store contract uses `{}` to mean “empty metadata only.”
> 
> The HTTP layer adds **`parseListSessionsQuery`** and **`foldDeepObjectQueryParam`** so Hono’s flat `metadata[env]=prod` keys are nested before Zod validation. **Bare `metadata=…` (JSON string) is rejected**, as are nested bracket forms, duplicates, and more than **10** filter keys. **Session metadata keys** are tightened with a regex (alphanumeric start; no `[]` or whitespace) so keys cannot collide with the bracket query syntax.
> 
> Filtering is wired through **`listSessions`** on in-memory, **Postgres** (`metadata @>` plus a new **GIN** migration), and **SQLite** stores. OpenAPI, Fern (`object-query-parameters`, `deepObjectMapQueryParameters`), and the TypeScript SDK expose the new **`metadata`** request field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b380592e46f14acb4c0a87a037d376dd19885550. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/fern/openapi/openapi.json
```diff
@@ -6850,6 +6850,24 @@
...
             }
+          },
+          {
+            "description": "Exact metadata pairs as metadata[key]=value. Sessions must contain all pairs.",
+            "explode": true,
+            "in": "query",
+            "name": "metadata",
+            "required": false,
+            "schema": {
+              "allOf": [
```
heerambavi1998: no allOf here please

packages/trueforge-core/src/agent-session/schemas/session.ts
```diff
@@ -19,6 +23,7 @@ export const SessionMetadataSchema = z
       .string()
       .min(1)
       .max(SESSION_METADATA_MAX_KEY_LENGTH)
+      .regex(SESSION_METADATA_KEY_PATTERN)
```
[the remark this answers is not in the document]
heerambavi1998: Can you check this?
Also add unit test for the regex, with expected pass and fail scenarios.

packages/trueforge-core/src/agent-session/store/ISessionStore.ts
```diff
@@ -71,6 +72,12 @@ export interface ListSessionsInput {
         agent_ids: readonly string[];
       }
     | undefined;
+  /**
+   * When non-empty, only sessions whose metadata contains all key/value pairs
+   * (exact string equality; extra session keys are allowed). `undefined` or `{}`
```
heerambavi1998: wouldn't { } mean sessions with no metadata or would it mean no sessions since there are no keys? This behaviour seems inconsistent for other filters. For example for array filter, [] returns [] and not all.

packages/trueforge/src/schemas/session.ts
```diff
@@ -64,6 +64,87 @@ const IsoTimestampQueryParam = z.iso
...
+
+/**
+ * Fold Hono's flat deepObject query keys into `{ metadata: { key: value } }`.
+ * Passes non-metadata keys through unchanged.
+ *
+ * - `metadata[key]=value` → equality filter
+ * - `metadata[key][…]=…` → rejected (only one bracket level is allowed)
+ * - bare `metadata=…` → rejected (no JSON-string dual support)
+ */
+export function foldListSessionsMetadataQuery(query: object): Record<string, unknown> {
```
heerambavi1998: why is this needed? Can we simplify this ?

packages/trueforge/src/db/sqlite/session-store/queries/sessions.ts
```diff
@@ -263,6 +263,14 @@ export async function listSessions(
     query = query.where('agent_id', '=', input.agent_id);
   }
   query = whereCreatedByOrAgentIds(query, input.created_by_or_agent_ids);
+  if (input.metadata !== undefined) {
+    for (const [key, value] of Object.entries(input.metadata)) {
```
heerambavi1998: can we instead take a structured key value pair list as input to list sessions

packages/trueforge/src/apis/sessions.ts
```diff
@@ -480,7 +480,7 @@ export function createSessionsRouter(deps: SessionsRouterDeps) {
   };
 
   const listSessionsHandler: RouteHandler<typeof listSessionsRoute> = async c => {
-    const query = c.req.valid('query');
+    const query = parseListSessionsQuery(honoQueriesToRecord(c.req.queries()));
```
heerambavi1998, line 486: why do we need to write our own parseListSessionsQuery method

packages/trueforge-core/src/agent-session/store/ISessionStore.ts
sr07asthana: Correct, updated the code so that {} mean sessions with no metadata rather than no sessions at all

packages/trueforge/src/schemas/session.ts
sr07asthana: OpenAPI says metadata is deepObject (metadata[env]=prod), but Hono does not nest that. c.req.queries() gives flat keys like "metadata[env]", not { metadata: { env: "prod" } }. The fold step builds the map Zod expects, and rejects bare metadata= / nested metadata[key][op].

packages/trueforge/src/apis/sessions.ts
sr07asthana: - Our URL on wire looks like /?metadata[env]=prod&limit=10
- What we want Zod to see:
  { metadata: { env: "prod" }, limit: "10" }
- What Hono actually gives:
   { "metadata[env]": "prod", limit: "10" } i.e. flat string keys, no nested metadata object.

c.req.valid('query') validates against the OpenAPI/Zod shape that expects a nested metadata map. It never gets that nest from Hono, so the filter doesn’t validate the way the schema describes.

So we fold first (metadata[env] → metadata.env), then run the same Zod schema.

packages/trueforge/src/db/sqlite/session-store/queries/sessions.ts
sr07asthana: What extra advantage will that give us?

heerambavi1998: removes the Object.entries conversion, and caller and callee both know the expected input

packages/trueforge/src/apis/sessions.ts
heerambavi1998: Lets refactor and keep these non session specific

packages/trueforge-core/src/agent-session/schemas/session.ts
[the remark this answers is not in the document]
heerambavi1998: Why are we allowing `-` and `:`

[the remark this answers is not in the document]
sr07asthana: -, : are general patterns used by teams for metadata
eg, "user:alice" should be a valid key

heerambavi1998 (approved):

PR #614, opened 2026-09-07
feat: Store schedule run errors

sr07asthana:
## Summary

## Changes

-

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a DB migration and expands the ScheduleRun API contract; failure-path persistence is new behavior but limited to schedule execution, not auth or payments.
> 
> **Overview**
> Schedule runs now carry an optional **`reason`** when execution fails, so API consumers can see why a run ended in `failed` instead of only the status.
> 
> A new `reason` column is added via Postgres/SQLite migrations and wired through the schedule store: **`updateRunStatus` only keeps a reason for `failed` runs** (otherwise it stores `null`). The cron dispatch loop and manual “run now” path both call **`scheduleRunFailureReason`**, which uses a non-empty `Error.message` or the fallback `"Schedule run failed"`, when hand-off to the agent turn fails.
> 
> **ScheduleRun** responses (OpenAPI, Zod schema, SDK types, and wire mapping) now include **`reason` as a required nullable string** on every run record.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 831bfda8e6c546da63173e503375467bf7fb3080. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -22,6 +22,11 @@ import type { ControlLoop } from './Controller';
  */
 export const DISPATCH_BATCH_LIMIT = 20;
 
+/** Human-readable failure detail for a `failed` schedule run. */
+export function scheduleRunFailureReason(error: unknown): string {
+  return error instanceof Error && error.message.trim() !== '' ? error.message : 'Schedule run failed';
```
heerambavi1998, line 27: can we ensure the reasons are human readable ? sending error message as is won't be helpful. 
Ok not to handle every scenario, but think about the ones which might be relevant to end user.
For example - Session creation failed because User does not have access to the agent anymore

packages/trueforge-sdk/src/api/types/ScheduleRun.ts
```diff
@@ -7,6 +7,8 @@ export interface ScheduleRun {
     createdBySubject: TrueForge.CreatedBySubject;
     id: string;
     name: string;
+    /** Failure detail when status is failed; otherwise null. */
```
heerambavi1998: comment not needed here

packages/trueforge/src/controller/scheduleDispatch.ts
sr07asthana: Schedule hand-off failures arrive as generic HTTP/SDK errors (message buried in the body or "Agent not found"), so we can’t cleanly persist a user-facing reason without unwrap/rewrite hacks.

should we introduce typed hand-off errors with final UX copy at the throw site or accept a thin boundary mapper that unwraps API bodies / remaps a few known cases.

heerambavi1998: Lets look at the error types and fix later

heerambavi1998 (approved):

PR #627, opened 2026-09-08
[Authz 6/6] Add internal endpoint for schedule execution w/ Trueforge Service Key

chiragjn:
> [!NOTE]
> **Medium Risk**
> Touches scheduled execution, a new service-key-protected internal route, and TrueFoundry agent token vending for background turns—operational misconfiguration of TRUEFORGE_API_KEY could block schedules.
> 
> **Overview**
> Scheduled runs no longer go through multi-step public SDK calls from the controller. The dispatch loop now POSTs once to **`POST /api/internal/schedules/runs/execute`** with **`TRUEFORGE_API_KEY`**, passing only `schedule_run_id`. The server loads the run, schedule, and agent, impersonates the **schedule creator** via `requestContextFromCreatedBySubject`, and starts the turn in-process (same path as manual run-now).
> 
> **TrueFoundry mode:** schedule turns resolve models/MCP with **agent-scoped vended tokens** (saved agent + creator subject); turn skills use a dedicated **`TrueFoundryAdminSkillStore`** (service API key). Interactive session turns still use the caller token where applicable.
> 
> Adds **`TRUEFORGE_API_KEY`** (required when `STANDALONE=false`, default `trueforge-standalone` locally), **`createApiKeyAuthMiddleware`**, schedule store **`getRunById`**, OpenAPI/SDK **`client.internal.schedules.executeRun`**, and env/docs updates for controller → server HTTP handoff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77fc508f41b44ce3911bfd2e9b97cb1193b0c89f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/.env.example
```diff
@@ -123,6 +123,10 @@ REDIS_URL=redis://localhost:6379
 ## Service in a split server/controller deployment.
 # SERVER_URL=http://localhost:8790
 
+## Service credential used only by the dedicated controller to execute persisted schedule runs.
```
heerambavi1998: lets remove the use case I think we have more use cases too like skills.

packages/trueforge/.env.example
```diff
@@ -123,6 +123,10 @@ REDIS_URL=redis://localhost:6379
...
+## Required when STANDALONE=false.
```
heerambavi1998: it should only be required in true foundry mode?

packages/trueforge/tests/unit/controller/scheduleDispatch.test.ts
```diff
@@ -68,72 +71,56 @@ function fakeStore(dispatchItem: ScheduleDispatchItem) {
...
+    expect(request).toHaveBeenCalledTimes(1);
+    const [url, init] = request.mock.calls[0] ?? [];
+    expect(String(url)).toBe('http://trueforge.internal:8790/api/internal/schedules/runs/execute');
+    expect(init?.method).toBe('POST');
+    expect(new Headers(init?.headers).get('authorization')).toBe('Bearer service-key');
+    expect(init?.body).toBe(JSON.stringify({ schedule_run_id: 'run-1' }));
   });
+});
 
-  it('does not call create-turn when the session already has a turn', async () => {
```
heerambavi1998, line 117: Lets still ensure we have these test scenarios- even if those ar removed to the API mock tests

packages/trueforge/src/main.ts
```diff
@@ -436,14 +523,41 @@ async function createServerRuntime<TTransaction>(persistence: ServerPersistence<
...
-        baseUrl: `http://localhost:${String(configuration.PORT)}`,
+        executeRun,
       })
     : undefined;
 
+  // wrappers that resolve the request context from the Hono context
+  // because handlers receive the Hono context
+  const resolveModelProviderStore = (c: Context, runAsAgent?: AgentRecord) => {
+    const requestContext = resolveRequestContext(c);
+    return resolveModelProviderStoreByRequestContext(requestContext, runAsAgent);
```
heerambavi1998: do we then need resolveModelProviderStoreByRequestContext methods at all? just wrapper should be enough?

packages/trueforge/src/main.ts
```diff
@@ -468,6 +582,8 @@ async function createServerRuntime<TTransaction>(persistence: ServerPersistence<
     oidcClient,
     authenticator,
     authorizer,
+    executeScheduleRun: executeRun,
+    scheduleExecutionApiKey: configuration.TRUEFORGE_API_KEY,
```
heerambavi1998: this looks off as it is oddly specific

packages/trueforge/src/controller.ts
```diff
@@ -41,6 +41,33 @@ export function createController<TTransaction>(params: {
...
+  return createControllerWithExecutor({
+    scheduleStore: params.scheduleStore,
+    withTransaction: params.withTransaction,
+    logger: params.logger,
+    executeRun: createHttpScheduleRunExecutor(params),
+  });
+}
+
+/** Controller colocated with the standalone server; schedule execution stays in-process. */
+export function createInProcessController<TTransaction>(params: {
```
heerambavi1998: can we just have one method that handles the mode specific logic

packages/trueforge/src/controller/scheduleDispatch.ts
```diff
@@ -127,6 +116,45 @@ export async function startScheduleRun(params: {
     input: [{ type: 'user.message', content: schedule.manifest.task }],
     previous_turn_id: 'none',
     userRef: schedule.created_by_subject.subject_id,
+    agent: named,
+  });
+}
+
+/** Loads trusted schedule context from a run id, then executes it idempotently. */
+export async function executeScheduleRun<TTransaction>(params: {
```
heerambavi1998: why is this in controller? if its anyways being passed to the control loop, keep it outside

packages/trueforge/src/app.ts
```diff
@@ -312,23 +317,22 @@ export function createServerApp<TTransaction>(deps: ServerDeps<TTransaction>) {
       authMiddleware,
     ),
   );
+  app.route(
+    '/api/internal/schedules',
+    withAuth(
+      createScheduleExecutionRouter({
+        executeScheduleRun: deps.executeScheduleRun,
```
heerambavi1998: this is internal implementation that will always be the same, passing as dep looks off

packages/trueforge/.env.example
chiragjn: No, in distributed mode, controller runs as a separate process / pod and needs to talk to server

packages/trueforge/src/main.ts
chiragjn: removed this

packages/trueforge/.env.example
heerambavi1998: What is the expected credentials in distributed mode in non truefoundry case

packages/trueforge/src/controller.ts
```diff
@@ -1,38 +1,38 @@
...
 
-function createScheduleApiClient(params: { baseUrl: string; tls: TlsOptions }): TrueForge {
-  const baseUrl = normalizeTlsUrl({ url: params.baseUrl, enabled: params.tls.enabled });
-  const fetchImpl = createTlsFetch(params.tls);
-  return new TrueForge({
-    baseUrl,
-    auth: false,
-    ...(fetchImpl !== undefined ? { fetch: fetchImpl } : {}),
+/** HTTP transport used by the dedicated controller process. */
+export function createHttpScheduleRunExecutor(params: {
```
heerambavi1998: can we move this executor definition to scheduleDispatch.ts? Essentially scheduleDispatchLoop shouldn't need to take executeRun as input param. 
was this done for tests? Is there any simpler way?

packages/trueforge/src/apis/schedules.ts
```diff
@@ -111,6 +116,29 @@ function isScheduleOwner(requestContext: Pick<RequestContext, 'subject'>, create
   return requestContext.subject.id === created_by_subject_id;
 }
 
+export function createScheduleExecutionRouter(deps: { executeScheduleRun: (scheduleRunId: string) => Promise<void> }) {
```
heerambavi1998: can you add a comment as to why this is a separate router - I guess its internal endpoint?

packages/trueforge/src/main.ts
```diff
@@ -214,14 +228,66 @@ function buildResolveAgentStore(options: {
...
+      });
   }
-  return c =>
-    new TrueFoundrySandboxProviderStore<TTransaction>({
-      context: resolveRequestContext(c),
+  return () => persistenceStore;
+}
+
+/** Executes a persisted schedule run in this process, as the schedule's creator and agent. */
+function buildExecuteScheduleRun<TTransaction>(options: {
```
heerambavi1998: can we simplify this? main file should not have this logic?

packages/trueforge/src/apis/schedules.ts
```diff
@@ -219,14 +247,7 @@ export function createSchedulesRouter<TTransaction>(deps: SchedulesRouterDeps<TT
...
 
     try {
-      await startScheduleRun({
-        item: { run, schedule },
-        sessions: deps.sessions,
-        agentStore: deps.resolveAgentStore(c),
-        startTurn: async turnParams => {
-          await startTurnInProcess({ ...turnParams, deps: deps.resolveTurnDeps(c, agent) });
-        },
-      });
```
heerambavi1998, line 239: I feel this was a simpler abstraction as it took store etc in deps. Lets keep it like this

packages/trueforge/src/controller.ts
chiragjn: We can, but the params are different - baseUrl, apiKey, tls these are not available / applicable in standalone mode

packages/trueforge/src/apis/schedules.ts
```diff
@@ -229,14 +337,14 @@ export function createSchedulesRouter<TTransaction>(deps: SchedulesRouterDeps<TT
...
+      c.set(
+        'request_context',
+        requestContextFromCreatedBySubject({
+          tenant_id: schedule.tenant_id,
+          created_by_subject: schedule.created_by_subject,
+        }),
+      );
```
chiragjn: probably not required

packages/trueforge/src/main.ts
```diff
@@ -147,15 +157,15 @@ function createServiceFoundryServerClient(
...
   const { persistenceStore, client } = options;
   if (client === undefined) {
     return () => persistenceStore;
   }
-  return c =>
+  return rc =>
     new TrueFoundrySkillStore<TTransaction>({
       client,
-      context: resolveRequestContext(c),
+      context: rc,
```
chiragjn: nit: rename the arg to requestContext

heerambavi1998 (approved):

PR #653, opened 2026-09-09
Fix: Use public base URL in Use In Code SDK snippets

bhaveshpatel640:
## Summary

Use In Code snippets used the request origin (in-cluster Service DNS). They now use `PUBLIC_BASE_URL` via `getPublicBaseUrl()`.

Closes AGE-2147

## Changes

- Code-snippet `baseUrl` from `getPublicBaseUrl()`
- Unit test asserts `base_url`

## How was this tested?

- `pnpm --filter @truefoundry/trueforge test -- tests/unit/apis/agents.test.ts`

## Checklist

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to URL strings in read-only snippet responses and an optional query override; no auth or data mutation paths.
> 
> **Overview**
> **Use In Code** snippets no longer embed the raw request origin (which could be in-cluster DNS). The code-snippets endpoint now accepts an optional **`base_url`** query parameter and returns a **public** SDK `baseUrl` in the payload.
> 
> When **`base_url`** is omitted, the server combines the request origin with the pathname from **`PUBLIC_BASE_URL`** (e.g. `https://sample.com` + path `/trueforge` → `https://sample.com/trueforge`). When the UI loads snippets, it passes the browser-resolved public URL via the updated internal **`getCodeSnippets`** SDK call.
> 
> OpenAPI, Zod schemas, and generated SDK types are aligned with the new query param and updated **`base_url`** descriptions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5c145cb9b348f18eb58376ff2d097397e5de8e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/src/apis/agents.ts
```diff
@@ -148,7 +149,7 @@ export function createAgentsRouter<TTransaction>(deps: AgentsRouterDeps<TTransac
       {
         data: buildAgentCodeSnippets({
           agentName: record.name,
-          baseUrl: new URL(c.req.url).origin,
+          baseUrl: getPublicBaseUrl(),
```
chiragjn: This throws if `PUBLIC_BASE_URL` is set to `''`
How are we dealing with that?

bhaveshpatel640: Fixed it using PUBLIC_BASE_URL and fall back to the request origin, when it's ''

chiragjn (approved):

packages/trueforge/src/apis/agents.ts
chiragjn: actually the way it is implemented will seem odd for SaaS tenants, the public url config is not tenant specific, so all tenants will see the same url - is that okay?

Only other solution I can think is to

```
new URL(<pathname-from-public-base-url>, new URL(c.req.url).origin).href
```

bhaveshpatel640: Can we take an optional baseUrl on the code-snippets request from the FE?

If it's omitted, we fall back to request origin + path from `PUBLIC_BASE_URL`. FE can then pass the real tenant/public host and avoid `*.svc.cluster.local`.

PR #742, opened 2026-09-15
Set automountServiceAccountToken as false for the once where it is no…

abhishekpradeep124:
> [!NOTE]
> **Low Risk**
> Helm-only pod spec hardening with a configurable override; risk is mainly if a deployment unexpectedly depended on the default mounted token for in-cluster API access.
> 
> **Overview**
> **Disables automatic mounting of the Kubernetes ServiceAccount token** on the TrueForge **server** and **controller** Deployments by setting `automountServiceAccountToken: false` by default and wiring it through a new chart value.
> 
> Pods still use the chart’s ServiceAccount for identity, but the projected API token is no longer mounted unless operators set `automountServiceAccountToken: true` in values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f99798caa1be52320f5c9a899c98f289794eb183. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chiragjn (approved):

charts/trueforge/values.yaml
```diff
@@ -16,6 +16,8 @@ serviceAccount:
   annotations: {}
   # If empty and create is true, a name is generated from the fullname.
   name: ""
+# Mount the ServiceAccount token into the server and controller pods.
+automountServiceAccountToken: false
```
chiragjn, line 20: Should this not be placed under `serviceAccount:` section?

PR #747, opened 2026-09-15
fix: map ServiceFoundry 401/403 errors to 500 for API key mismatches

thesujai:
## Summary

AGE-2182

## Changes

-

## Checklist

---

> [!NOTE]
> **Low Risk**
> Narrow change in `#requestJson` error mapping for service-key requests only; user-token auth behavior is unchanged.
> 
> **Overview**
> When **ServiceFoundry** rejects a request with **401/403** and the call used the server **`TRUEFOUNDRY_API_KEY`** (not the user bearer), **TrueFoundry** now responds with **500** instead of forwarding **401/403**.
> 
> That keeps a bad or mismatched service key from being treated like an expired user session, which was **logging users out of the UI**. User-authenticated SFY calls still surface **401/403** unchanged (e.g. **`getSession`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b73cfa6c1209f0741d4c5ffbc3734afba0df5771. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


thesujai:
On second thought I dont think we need this PR at all, because in helm we derive the TRUEFOUNDRY_API_KEY from the same place for trueforge as well as serveicfoundry or other services. So misconfig of that is a bigger failure which will not even start the control plane

PR #765, opened 2026-09-16
feat: add Redis Sentinel and TLS support in TrueForge

thesujai:
## Summary

feat: add Redis Sentinel and TLS support in TrueForge

Closes AGE-2071

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how every peered replica connects to Redis and how Helm injects Redis config; misconfiguration or Sentinel/TLS edge cases could block startup or break executor peering until fixed.
> 
> **Overview**
> Distributed TrueForge can now connect to Redis via **URL**, **host/port + auth**, **Redis Sentinel**, and optional **TLS/mTLS**, instead of relying on a single `REDIS_URL` with a localhost default.
> 
> The server reads a broader **`REDIS_*` env surface** (host, port, DB, credentials, TLS material, Sentinel nodes/master/auth). **`connectRedis`** chooses Sentinel vs URL vs host mode, applies shared TLS socket options, loads PEM from paths or inline text, and **rebuilds Sentinel clients on connect failure** (except auth errors). Request-reply and event streaming use a shared **`RedisPeerClient`** type; **Sentinel reuses one client for pub/sub** (no `duplicate()` subscriber), with shutdown adjusted accordingly.
> 
> The **Helm chart** maps expanded **`externalRedis`** (url/host/auth/tls/sentinel) into those env vars, adds render-time validation (bundled vs external mutual exclusion, required fields), and documents upgrade-friendly **`externalRedis.enabled`** behavior. **`.env.example`** and chart README/NOTES are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49bd3b18a0b7052b534502bd31e0f8bafec15765. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge/.env.example
```diff
@@ -56,9 +56,33 @@ PORT=8790
 # SQLITE_PATH=
 
 ## Used when STANDALONE=false (ignored in standalone). Redis connection URL for
-## executor peering. Defaults to redis://localhost:6379 if unset.
+## executor peering. Preferred over REDIS_HOST when set (may include userinfo).
```
sr07asthana: Trying to understand the reasoning behind Preferred over REDIS_HOST when set.

packages/trueforge/src/config.ts
sr07asthana, line 1: Instead of accepting all env var in config and silently overwriting based on priority, i think we should validate and fail fast in case of incorrect config.
1. standalone=false At least one of: Sentinel complete / URL / host should be set
2. reject overlapping transports (URL+host, or Sentinel+URL/host) instead of silent priority
3. If REDIS_SENTINEL_ENABLED=true, require nodes + master name

check how we're handling postgres i think we're failing fast at startup.

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,285 @@
...
+  caCert: string | undefined;
+  rejectUnauthorized: boolean | undefined;
+  serverName: string | undefined;
+  cert: string | undefined;
+  key: string | undefined;
+  keyPassphrase: string | undefined;
+}
+
+export function isStandaloneRedisClient(client: RedisPeerClient): client is RedisClientType {
+  return 'duplicate' in client;
```
sr07asthana: Is there a better way to figure out if the client isStandaloneRedisClient

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,285 @@
...
 import { extractErrorLogFields } from '@truefoundry/trueforge-core/core';
-import { createClient, type RedisClientType } from 'redis';
+import type { RedisPeerClient } from '@truefoundry/trueforge-core/request-reply';
+import { createClient, createSentinel, type RedisClientType } from 'redis';
 import type { Logger } from 'winston';
 
-export async function connectRedis(input: { url: string; logger: Logger }): Promise<RedisClientType> {
+const DEFAULT_SENTINEL_PORT = 26379;
+const DEFAULT_REDIS_PORT = 6379;
+const DEFAULT_REDIS_DB = 0;
```
sr07asthana: shouldn't these be default values in config?

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,285 @@
...
+  /** Host / Sentinel mode only. Defaults to 0. */
+  database: number | undefined;
+  username: string | undefined;
+  password: string | undefined;
+  logger: Logger;
+  sentinel:
+    | {
+        enabled: boolean | undefined;
+        nodes: string | undefined;
+        masterName: string | undefined;
```
sr07asthana: here as well, having a flat input doesn't seem to be correct design
maybe we can have a discriminated union like
type RedisConnection =
  | { mode: 'url'; url: string }
  | {
      mode: 'host';
      host: string;
      port?: number;
      database?: number;
      username?: string;
      password?: string;
    }
  | {
      mode: 'sentinel';
      nodes: string;
      masterName: string;
      database?: number;
      /** Auth for Redis data nodes (master/replicas). */
      username?: string;
      password?: string;
      /** Auth for Sentinel processes. */
      sentinelUsername?: string;
      sentinelPassword?: string;
    };
type ConnectRedisInput = {
  connection: RedisConnection;
  tls: RedisTlsInput | undefined;
  logger: Logger;
};

we can also put validation on config as well in this case.

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,285 @@
...
+    // createSentinel()'s return is not assignable to RedisSentinelType under exactOptionalPropertyTypes
+    // @ts-expect-error TS2375
+    client = await connectSentinelWithRetry({
+      sentinel: input.sentinel,
+      auth,
+      clientDefaults,
+      socketTls,
+      logger: input.logger,
+    });
+  } else if (url) {
```
sr07asthana: this branching will become simpler with discriminated union

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,285 @@
...
+    disableOfflineQueue: true,
+    pingInterval: PING_INTERVAL_MS,
+  } as const;
+  const url = input.url?.trim();
+
+  let client: RedisPeerClient;
+  if (isRedisSentinelConfigured(input.sentinel) && input.sentinel) {
+    // createSentinel()'s return is not assignable to RedisSentinelType under exactOptionalPropertyTypes
+    // @ts-expect-error TS2375
+    client = await connectSentinelWithRetry({
```
sr07asthana: will this keep on trying to reconnect forever? Until Sentinel connect() succeeds, will startup be stuck in that retry loop 
cc @chiragjn what should be the expected approach here?

sr07asthana: in standalone=false mode, for our processes like peering and turn event streaming since redis is a hard dependency, we should ensure it is up a server boot. But lets keep MAX_RETRIES or MAX_TIME it can retry upto and then fail?

thesujai: This is how gateway does this, inifinite retries

ramantehlan (approved):

charts/trueforge/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: TrueForge server (API + UI) served from a single container image.
 type: application
 # version / appVersion are maintained on main (bot chart-release PR or human).
 # Publishing is gated by git tag charts/trueforge@<version> (must match version).
-version: "0.2.2"
+version: "0.2.3"
```
chiragjn, line 7: minor: 
no need to bump here

packages/trueforge/src/config.ts
```diff
@@ -349,13 +363,92 @@ function resolveCodeModeSocketParent(): string {
...
+      defaultValue: 6379,
+    });
+    return {
+      mode: 'host',
+      host,
+      port,
+      database,
+      ...(username !== undefined ? { username } : {}),
+      ...(passwordOrUndefined !== undefined ? { password: passwordOrUndefined } : {}),
+    };
```
chiragjn, line 477: we should return `{mode: 'standalone', url: ...}`
url can be made from components

I am not sure about the tls part
redis:// vs rediss://
when making this url

charts/trueforge/templates/_helpers.tpl
```diff
@@ -433,20 +433,87 @@ fields, wires bundled Postgres/Redis, optional OIDC, then server.extraEnv.
...
 {{- $env = append $env (dict "name" "GRACEFUL_TIMEOUT_SECONDS" "value" (.Values.server.gracefulTimeoutSeconds | toString)) -}}
 
+{{- $externalRedis := .Values.externalRedis | default dict -}}
+{{- $sentinel := $externalRedis.sentinel | default dict -}}
+{{- $auth := $externalRedis.auth | default dict -}}
+{{- $tls := $externalRedis.tls | default dict -}}
+{{- $sentinelEnabled := eq (toString ($sentinel.enabled | default false)) "true" -}}
+{{- $tlsEnabled := eq (toString ($tls.enabled | default false)) "true" -}}
+{{- $externalRedisActive := or $externalRedis.enabled $externalRedis.url $externalRedis.host $sentinelEnabled -}}
+{{- if and .Values.redis.enabled $externalRedisActive -}}
```
chiragjn: why is $externalRedis.enabled not enough here?

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,274 @@
...
+    ...(tls.cert ? { cert: resolvePemMaterial({ value: tls.cert, label: 'REDIS_TLS_CERT' }) } : {}),
+    ...(tls.key ? { key: resolvePemMaterial({ value: tls.key, label: 'REDIS_TLS_KEY' }) } : {}),
+    ...(tls.keyPassphrase ? { passphrase: tls.keyPassphrase } : {}),
+    ...(tls.serverName ? { servername: tls.serverName } : {}),
+  };
+}
+
+function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
```
chiragjn: nit: pretty sure this is duplicated

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,274 @@
...
+      passphrase?: string;
+      servername?: string;
+    }
+  | undefined {
+  if (!tls?.enabled) {
+    return undefined;
+  }
+  if ((tls.cert && !tls.key) || (tls.key && !tls.cert)) {
+    throw new Error(
+      '[Redis] mTLS misconfigured: REDIS_TLS_CERT and REDIS_TLS_KEY must be set together ' +
```
chiragjn:
```suggestion
      '[Redis] TLS misconfigured: REDIS_TLS_CERT and REDIS_TLS_KEY must be set together ' +
```

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,274 @@
...
+function buildTlsSocketOptions(tls: RedisTlsInput | undefined):
+  | {
+      tls: true;
+      rejectUnauthorized: boolean;
+      ca?: string;
+      cert?: string;
+      key?: string;
+      passphrase?: string;
+      servername?: string;
+    }
```
chiragjn: nit: repeating enough to be a type of its own

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,274 @@
...
+      database: number;
+      username?: string;
+      password?: string;
+      sentinelUsername?: string;
+      sentinelPassword?: string;
+    };
+
+export type ConnectedRedis =
+  | { mode: 'url'; client: RedisClientType }
+  | { mode: 'host'; client: RedisClientType }
```
chiragjn:
```suggestion
  | { mode: 'standalone'; client: RedisClientType }
```

packages/trueforge/src/runtime/redis.ts
```diff
@@ -1,21 +1,274 @@
...
+  | { mode: 'sentinel'; client: RedisPeerClient };
```
chiragjn:
```suggestion
  | { mode: 'sentinel'; client: RedisSentinelType };
```

packages/trueforge-core/src/request-reply/redisClient.ts
```diff
@@ -0,0 +1,4 @@
+import type { RedisClientType, RedisSentinelType } from 'redis';
+
+/** Standalone or Sentinel client for request-reply / command use. */
+export type RedisPeerClient = RedisClientType | RedisSentinelType;
```
chiragjn:
```suggestion
export type RedisClient = RedisClientType | RedisSentinelType;
```

charts/trueforge/Chart.yaml
ramantehlan: We need to release trueforge chart 0.2.3, since truefoundry helm chart changes depends on these features going live.

packages/trueforge/src/config.ts
thesujai: Done — host/url collapsed to `{ mode: 'standalone', url }`. Host fields build a `redis://` URL; TLS stays on `REDIS_TLS_*` socket options (CA/cert/key cannot live in the URL scheme).

charts/trueforge/templates/_helpers.tpl
thesujai: Agreed — gated on `externalRedis.enabled` only now (docs/values updated).

packages/trueforge/src/runtime/redis.ts
thesujai: Removed the local `sleep`; using `node:timers/promises` instead.

packages/trueforge/src/runtime/redis.ts
thesujai: Updated to "TLS misconfigured".

packages/trueforge/src/runtime/redis.ts
thesujai: Extracted `RedisTlsSocketOptions`.

packages/trueforge/src/runtime/redis.ts
thesujai: `ConnectedRedis` is now `standalone | sentinel`.

packages/trueforge/src/runtime/redis.ts
thesujai: Sentinel arm now uses `RedisSentinelType`.

packages/trueforge-core/src/request-reply/redisClient.ts
thesujai: Renamed to `RedisClient`.

charts/trueforge/Chart.yaml
thesujai: Leaving the 0.2.3 bump per @ramantehlan — needed for the parent chart release.

charts/trueforge/templates/_helpers.tpl
ramantehlan: As discussed, taking missing enabled as false now.

packages/trueforge/src/config.ts
```diff
@@ -349,13 +363,120 @@ function resolveCodeModeSocketParent(): string {
...
+    auth = encodeURIComponent(parts.username);
+    if (parts.password !== undefined) {
+      auth += `:${encodeURIComponent(parts.password)}`;
+    }
+    auth += '@';
+  } else if (parts.password !== undefined) {
+    auth = `:${encodeURIComponent(parts.password)}@`;
   }
-  return raw;
+  const dbPath = parts.database > 0 ? `/${String(parts.database)}` : '';
```
chiragjn: any harm in always appending `/{parts.database}` ?

packages/trueforge/src/config.ts
```diff
@@ -349,13 +363,120 @@ function resolveCodeModeSocketParent(): string {
...
+  let auth = '';
+  if (parts.username !== undefined) {
+    auth = encodeURIComponent(parts.username);
+    if (parts.password !== undefined) {
+      auth += `:${encodeURIComponent(parts.password)}`;
+    }
+    auth += '@';
+  } else if (parts.password !== undefined) {
+    auth = `:${encodeURIComponent(parts.password)}@`;
   }
```
chiragjn, line 398: can we use 

url = new URL('redis://host:port/database')
url.username = 
url.password = 
url.href

does url.href also automatically take care of encoding?

packages/trueforge/src/config.ts
```diff
@@ -349,13 +363,120 @@ function resolveCodeModeSocketParent(): string {
...
-  if (raw.trim() === '') {
-    throw new Error('Environment variable REDIS_URL must be non-empty when STANDALONE=false.');
+/** Non-empty trimmed env string, or undefined. */
+function optionalNonEmptyEnv(envKey: string): string | undefined {
+  const raw = getEnv(envKey);
+  if (raw === undefined || raw.trim() === '') {
+    return undefined;
+  }
+  return raw.trim();
+}
```
chiragjn, line 374: rather not put any intelligence here, let people fix extra spaces in their values

just string or undefined

packages/trueforge/src/main.ts
```diff
@@ -779,12 +815,15 @@ try {
...
       // Stop serving peer requests (waits for in-flight replies), then close
-      // the clients this process owns: the subscriber duplicate and the primary.
+      // clients this process owns: the subscriber duplicate (standalone only)
+      // and the primary.
       await requestReplyExecutor?.drain();
-      await requestReplySubscriber?.close().catch((error: unknown) => {
-        logger.warn('[Redis] Error closing subscriber client during shutdown', extractErrorLogFields(error));
-      });
-      await redis?.close().catch((error: unknown) => {
+      if (requestReplySubscriberOwned) {
```
chiragjn: if close is idempotent, skip this `if`

chiragjn (approved):
Keeping it approved, I will check the helpers.tpl separately

PR #774, opened 2026-09-16
fix: don't close a shared MCP transport under in-flight sibling calls

bhaveshpatel640:
## Summary

Session-expired MCP retry no longer closes the shared transport while a sibling `callTool`/`listTools` is still using it. That sibling used to fail with a non-retried close error.

AGE-2253 / INFOSEC-15 TF-06

## Changes

- Detach the stale socket (`pendingClose`) and close it when `inflight` hits 0
- Skip reset if the expired handle is no longer the live connection

## How was this tested?

`pnpm --filter @truefoundry/trueforge-core test -- tests/core/mcp/remoteMcpServer.test.ts`

## Checklist

- [x] No hand-edits to generated code (`packages/trueforge-sdk`, `python/trueforge_sdk`, `.github/fern/openapi/openapi.json`, `docs/openapi.json`) — fork PRs omit SDK regen; maintainers regenerate after merge

---

> [!NOTE]
> **Medium Risk**
> Changes concurrent connection lifecycle and session-expiry retry in remote MCP tooling; behavior is localized but affects reliability of parallel agent tool calls.
> 
> **Overview**
> Fixes a race where **session-expired retry** on one remote MCP `callTool`/`listTools` could **close the shared transport** while another operation was still using it. The sibling then failed with **"Connection closed"**, which is not treated as session-expired, so it was **never retried**.
> 
> `RemoteMCP` now tracks **`inflight`** operations and **`pendingClose`**: on expiry it **detaches** the stale socket instead of closing it immediately, **reconnects and retries once**, and **closes `pendingClose` only when `inflight` hits 0**. **`resetConnection`** ignores expiry on a connection that is no longer the live one so a late failure on an old socket cannot tear down a newer transport. Transport **`onClose`** only clears connected state when the closed handle is still current.
> 
> Tests cover overlapping calls (hold/release), deferred close counts, and stale sibling expiry after a successful retry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb590da7d5cc8e31a8e22dbf3688bfe0832e660a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/trueforge-core/src/core/mcp/RemoteMCP.ts
```diff
@@ -248,6 +294,7 @@ export class RemoteMCP implements ToolSource {
         connection = await this.tracing.withRemoteMcpToolSpan(
           { method: 'initialize', serverName: this.name, serverId: this.id, serverUrl: this.traceUrl, enabled: true },
           async span => {
+            const attached: { current: RemoteMcpConnection | undefined } = { current: undefined };
```
chiragjn: simplify to just `currentConnection` ?

chiragjn (approved):

packages/trueforge-core/src/core/mcp/RemoteMCP.ts
bhaveshpatel640: Done

PR #778, opened 2026-09-17
feat: allow user message when turn is running

sr07asthana:
## Summary

## Changes

-

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core turn send validation, tool-call repair, sub-agent lifecycle, and how thread snapshots are merged across turns—behavioral regressions could affect resumption and persisted conversation state.
> 
> **Overview**
> Users can **start a new turn with a user message** even when approvals, client-side tools, or open sub-agent threads are still pending, instead of being blocked or forced to finish those interactions first.
> 
> **Runtime steering:** `OpenToolCallCloser` now receives a `userMessageIncoming` flag on `preSend`. For resume batches it still only auto-closes dangling regular tool calls; when a user message arrives it synthesizes cancelled tool responses for all unmatched calls on the last assistant message (including approval, client-side, and thread-creation calls) so the new user content can follow valid tool pairs. `AgentThread` no longer rejects user sends for pending approvals; empty sends still fail until pending work is resolved unless a user message interrupts. The orchestrator **drops in-memory child threads** before routing a non-approval user batch to the root and removes the old error that forbade user messages while sub-agents were active.
> 
> **Persistence:** Turn snapshots treat `capability_states` as the authoritative **live thread set**—threads omitted from it are not copied from the previous turn (in-memory, Postgres, and SQLite stores). Validation now requires context appends to target live threads only.
> 
> Also fixes root `AGENT_DONE` terminal handling so **`output` is attached only when status is `done`**, uses `InvalidAgentSendInputError` for homogeneous send-batch validation, and adds orchestration/session/store tests plus an SDK regen changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed91f6030b205a50afda915548f39e386578edc8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/fern/openapi/openapi.json
```diff
@@ -4575,9 +4576,27 @@
...
+            "enum": [
+              "cancelled"
+            ],
+            "type": "string"
+          }
+        },
+        "required": [
+          "status"
+        ],
+        "type": "object"
```
heerambavi1998, line 4598: why is every thread state object? isn't it just an enum ?

packages/trueforge-core/src/core/runtime/AgentThread.ts
```diff
@@ -316,11 +296,9 @@ function validateInputMessageTypesGivenContext(
   context: ContextMessage[],
   messages: AgentThreadRuntimeSendInput[],
 ): void {
+  const hasUserMessage = messages.some(isInputUserMessage);
   // Full open set: validates incoming tool responses and dedupes within the batch.
   const openToolCallIds = getOpenToolCallIds(context);
-  // Subset that blocks a fresh user message: excludes calls OpenToolCallCloser will auto-close
```
heerambavi1998, line 321: why was this removed?

packages/trueforge-core/src/core/runtime/AgentThread.ts
```diff
@@ -347,9 +324,9 @@ function validateInputMessageTypesGivenContext(
     }
   }
 
-  // A send for a thread awaiting user input must resolve every pending approval and client-side
-  // tool call in the same batch; any left unresolved (including an empty batch) is a blocker.
-  if (pendingApprovalIds.size > 0 || pendingClientSideIds.size > 0) {
+  // User messages interrupt pending approvals / client-side calls (OpenToolCallCloser
+  // synthesizes the missing responses). Empty and action batches still must resolve them all.
+  if (!hasUserMessage && (pendingApprovalIds.size > 0 || pendingClientSideIds.size > 0)) {
```
heerambavi1998: can we short circuit the hasUserMessage so we don't have to add a check here

packages/trueforge-core/src/core/runtime/AgentThread.ts
```diff
@@ -503,6 +480,7 @@ export class AgentThread {
   private deferredTool?: DeferredTool | undefined;
   private convertedTools: ConvertToolsResult | undefined;
   private pendingSandboxCreatedEvents: SandboxCreatedEvent[] = [];
+  private pendingPreSendOutputEvents: ToolResponseEvent[] = [];
```
heerambavi1998, line 485: can you explain what this is for

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
```diff
@@ -45,7 +45,8 @@ function agentThreadEventToTerminalFields(event: AgentThreadExecutionEvent): {
       if ('parent' in event && event.parent) {
         return {};
       }
-      if (event.status === 'error') {
+      // only a successful root completion becomes turn output
+      if (event.status !== 'done') {
         return {};
       }
       return { output: event.output };
```
heerambavi1998: lets do if (event.status === 'done' } return output else { }

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
```diff
@@ -165,7 +166,7 @@ function createRootAgentSpan(mainThread: AgentThread, tracing: AgentTracing): Ro
         if (isInternalThreadDoneError(chunk)) {
           rootAgentErrorMessage = chunk.error;
           trace.setOutput(JSON.stringify({ error: chunk.error }));
-        } else {
+        } else if (chunk.status === 'done') {
```
heerambavi1998: what about cancelled? What are we setting on span.

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
```diff
@@ -207,7 +208,7 @@ export async function* wrapWithSubAgentSpan(
           subTrace.setOutput(JSON.stringify({ error: event.error }));
           subTrace.setMetrics(currentThread.getAgentThreadMetrics());
           subTrace.setError(event.error);
-        } else {
+        } else if (event.status === 'done') {
```
heerambavi1998: same here add explicit cancel handling

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
```diff
@@ -261,9 +263,24 @@ export class AgentThreadOrchestrator {
...
+  private markNonRootThreadsCancelled(): void {
+    for (const thread of this.agentThreads.values()) {
+      if (thread.parent) {
+        this.cancelledThreadIds.add(thread.threadId);
+      }
+    }
+  }
+
   public async *send(messages: AgentThreadSendBatch): AsyncGenerator<AgentThreadAppendContext, void, unknown> {
+    if (messages.length > 0 && !isUserToolApprovalOrResponseBatch(messages)) {
```
heerambavi1998, line 282: We should have a separate CancelTurn method that does all this ( that we later reuse in cancel API as well) 
We shouldn't add such conditional handling directly in orchestrator. 
Perhaps we can model it as a "send cancel" event to orchestrator? @chiragjn thoughts here?

.github/fern/openapi/openapi.json
sr07asthana: They carry fields apart from status only
- ThreadStateError carries error field, which is human readable error
- ThreadStateError and ThreadStateDone carry output field, which carries thread output that need to be appended to main threads context
- ThreadStateCancelled will not have any output field

packages/trueforge-core/src/core/runtime/AgentThread.ts
sr07asthana: It existed to answer basically which open tool calls would still block a user message after the open tool call closer runs. Closer used to auto-close only dangling regular calls. Approval, client-side, and create_sub_agent stayed open. So validation did: blocking = open − closable-dangling

We made a user message close all unmatched last-assistant calls. Nothing is “unclosable” on that path anymore.

packages/trueforge-core/src/core/runtime/AgentThread.ts
sr07asthana: createTurn only drains send() for context (collectContextAppends ignores output). If we yielded those events from send(), they would never be persisted.
So this:
Copies tool.response items into pendingPreSendOutputEvents.
Yields the append with output: [] so context still lands in the snapshot.
and when we run execute we flush pendingPreSendOutputEvents

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
sr07asthana: Added

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
sr07asthana:
```
else if (isInternalThreadDoneCancelled(chunk)) {
          rootAgentErrorMessage = 'Agent thread cancelled';
          trace.setOutput(JSON.stringify({ cancelled: true }));
        }
```
added this handling for cancelled case.

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
sr07asthana: cancel api is separate in our case right? we decided cancel and closing these calls will be two separate processess right?
The previous turn is already frozen. You start a new turn with a real user message while a sub-agent (or approval) was still open in the snapshot.
Then this orchestrator flow runs, in send() of that new turn

We discussed yesterday that cancel itself should not close all pending subagent/pending actions. It should be done when next user message arrives. cc @chiragjn

packages/trueforge-core/src/core/runtime/AgentThread.ts
heerambavi1998: but do we need to persist?

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
heerambavi1998: I did not get this.
A user message creates a new turn, which should cancel the previous turn. So sending a user message should have the exact same behaviour as say clicking cancel turn button. 
This was what we discussed. 

The problem here is that we already mark the turn as done. I believe this change will make more sense once we pause the turn instead. 
Currently the only scenario we need to handle is that the turn is running, and we send a user message - which interrupts the turn.

packages/trueforge-core/src/core/runtime/AgentThread.ts
sr07asthana: suppose client already received HITL required actions, on continuing turn we simply close them on BE without yielding and persisting tool closure events, how will client know this was closed synthetically?
we need this so the session event log and UI to show those calls as closed

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
sr07asthana: freeze/cancel-button stops the old turn, this send() path stops leftover threads/tools on the next turn.

packages/trueforge-core/src/core/runtime/AgentThreadOrchestrator.ts
```diff
@@ -261,7 +261,20 @@ export class AgentThreadOrchestrator {
...
+        addAgentThreadMetrics(this.finishedSubAgentMetrics, thread.getAgentThreadMetrics());
+        this.agentThreads.delete(thread.threadId);
+      }
+    }
+  }
+
   public async *send(messages: AgentThreadSendBatch): AsyncGenerator<AgentThreadAppendContext, void, unknown> {
+    if (messages.length > 0 && !isUserToolApprovalOrResponseBatch(messages)) {
+      this.dropNonRootThreads();
+    }
```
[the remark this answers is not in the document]
sr07asthana: This is accepted, for now we have decided it will be clients responsibility to handle previous pending tool calls, user actions and threads in case a new user message is sent

packages/trueforge-core/src/core/runtime/OpenToolCallCloser.ts
```diff
@@ -8,78 +8,103 @@
...
       return;
     }
 
-    const dummyToolMessages: LLMToolMessage[] = [...requestedIds].map(toolCallId => ({
+    const content = options.userMessageIncoming ? CANCELLED_TOOL_MESSAGE_CONTENT : DANGLING_TOOL_MESSAGE_CONTENT;
+    const dummyToolMessages: LLMToolMessage[] = closableIds.map(toolCallId => ({
       role: 'tool',
       tool_call_id: toolCallId,
-      content: DUMMY_TOOL_MESSAGE_CONTENT,
+      content,
```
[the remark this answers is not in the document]
sr07asthana: This is accepted, for now we have decided it will be clients responsibility to handle previous pending tool calls, user actions and threads in case a new user message is sent

packages/trueforge-core/src/agent-session/store/ISessionStore.ts
```diff
@@ -117,7 +117,7 @@ export interface TurnContextAppend {
 
 export interface CreateTurnInput<TTurnCustom extends object = Record<string, never>> {
   turn: TurnRecordWithoutSnapshot<TTurnCustom>;
-  /** Threads absent on the previous turn (turn 1: the root thread; else empty). */
+  /** Live threads absent on the previous turn (turn 1: the root thread; else usually empty). */
```
debajyoti-truefoundry, line 120:
```suggestion
  /** Threads absent on the previous turn (turn 1: the root thread; else empty). */
```

